PR #1, opened 2015-06-22
Move over Diff code from DataModel

JeroenDeDauw:
This is a move with of the DataModel Diff code with minimal changes.
The namespace has been changed from `Wikibase\DataModel\Entity\Diff`
to `Wikibase\DataModel\Services\Diff`. Furthermore, the package
private classes have been put into a `Internal` sub namespace to
make the visibility more distinct.

This change allows for removing the Diff code, and the dependency
on the Diff library, from Wikibase DataModel in its next major
release.

The objects representing the diffs themselves contain some technical
debt. With this move we can solve this when we see fit without
having to create a breaking release for Wikibase DataModel itself.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
 	},
 	"require": {
 		"php": ">=5.3.0",
-		"wikibase/data-model": "~3.0"
+		"wikibase/data-model": "~3.0",
+		"diff/diff": "~2.0"
 	},
 	"require-dev": {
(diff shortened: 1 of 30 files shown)


JeroenDeDauw:
ScrutinizerCI failure is unrelated. The thing is not properly configured yet for some reason...

JeroenDeDauw:
The ScrutinizerCI failure is unrelated.

PR #2, opened 2015-06-28
Move StatementGuidGenerator code from WB/Lib

addshore:


src/Statement/GuidGenerator.php
```diff
@@ -0,0 +1,40 @@
...
+	 * @var V4GuidGenerator
+	 */
+	private $baseGenerator;
+
+	public function __construct() {
+		$this->baseGenerator = new V4GuidGenerator();
+	}
+
+	/**
+	 * Generates and returns a GUID for a claim in the given Entity.
```
Benestar: claim -> statement

PR #5, opened 2015-06-30
[WIP] Add validators from wikibase/lib

Benestar:
This introduces dependencies on data-values components. If we don't
want them in here we have to introduce even another component.
EntityLookup should be moved here in another patch.
Parser::EXT_LINK_URL_CLASS got replaced with a private property
and should be set in the constructor to the constant defined in Parser.

**WIP**: Tests missing; do we want/need to somehow import the git history of the moved code?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,9 @@
 	"require": {
 		"php": ">=5.3.0",
 		"wikibase/data-model": "~3.0",
-		"diff/diff": "~2.0"
+		"diff/diff": "~2.0",
+		"data-values/data-values": "~1.0",
+		"data-values/interfaces": "~0.1"
 	},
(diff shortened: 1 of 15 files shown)


JeroenDeDauw:
As discussed in person with @Benestar, a lot of this code does not belong in this component, since it does not bind to the DataModel in any way.

thiemowmde:
Uh, I did not expected this, @Benestar. I have to agree to @JeroenDeDauw. Only 2 classes use stuff from DataModel.

addshore:
Closing as all of the Validators stuff has been moved to the Repo component now!

https://gerrit.wikimedia.org/r/229742
https://gerrit.wikimedia.org/r/229744

PR #10, opened 2015-07-28
Remove BestStatementsFinder as it is unused

Benestar:


JonasKress:
This could be useful for the lua extension? Keep this and extend it towards being able to get only statements that have a reference?

Benestar:
I should've repeated what I wrote in the [other pull request](https://github.com/wmde/WikibaseDataModel/pull/516#issuecomment-124976242): This can easily be done using `StatementList` with `StatementList::getByProperty::getBestStatements`. What this class provides is quite redundant and should be removed.

Concerning references, we should discuss that before implementing anything since not all references are valid on all Wikipedias.

PR #16, opened 2015-08-07
Add basic lookups from WikibaseLib

addshore:
This is a CP from Wikibase lib with 1 change

EntityLookup now throws OutOfBoundsException
insead of a StorageException!!!!!!!!!!!!!

### DIFF
--- a/src/Lookup/EntityLookup.php
+++ b/src/Lookup/EntityLookup.php
@@ -2,7 +2,6 @@
 
 namespace Wikibase\DataModel\Services\Lookup;
 
-use OutOfBoundsException;
 use Wikibase\DataModel\Entity\EntityDocument;
 use Wikibase\DataModel\Entity\EntityId;
 
@@ -25,11 +24,10 @@ interface EntityLookup {
 	 * Code that needs control over redirect resolution should use an
 	 * EntityRevisionLookup instead.
 	 *
-	 * @since 0.3
+	 * @since 1.1
 	 *
 	 * @param EntityId $entityId
 	 *
-	 * @throws OutOfBoundsException
 	 * @return EntityDocument|null
 	 */
 	public function getEntity( EntityId $entityId );
@@ -43,11 +41,10 @@ interface EntityLookup {
 	 * Code that needs control over redirect resolution should use an
 	 * EntityRevisionLookup instead.
 	 *
-	 * @since 0.4
+	 * @since 1.1
 	 *
 	 * @param EntityId $entityId
 	 *
-	 * @throws OutOfBoundsException
 	 * @return bool
 	 */
 	public function hasEntity( EntityId $entityId );
--- a/src/Lookup/EntityLookup.php
+++ b/src/Lookup/EntityLookup.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Lookup;
+
+use OutOfBoundsException;
+use Wikibase\DataModel\Entity\EntityDocument;
+use Wikibase\DataModel\Entity\EntityId;
+
+/**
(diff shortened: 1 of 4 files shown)


src/Lookup/EntityLookup.php
```diff
@@ -0,0 +1,55 @@
...
+	 *
+	 * @note Implementations of this method may or may not resolve redirects.
+	 * Code that needs control over redirect resolution should use an
+	 * EntityRevisionLookup instead.
+	 *
+	 * @since 0.4
+	 *
+	 * @param EntityId $entityId
+	 *
+	 * @throws OutOfBoundsException
```
JeroenDeDauw: Why is there a @throws for this method?

src/Lookup/EntityLookup.php
```diff
@@ -0,0 +1,55 @@
...
+	/**
+	 * Returns whether the given entity can bee looked up using
+	 * getEntity(). This avoids loading and deserializing entity content
+	 * just to check whether the entity exists.
+	 *
+	 * @note Implementations of this method may or may not resolve redirects.
+	 * Code that needs control over redirect resolution should use an
+	 * EntityRevisionLookup instead.
+	 *
+	 * @since 0.4
```
JeroenDeDauw: .

src/Lookup/EntityLookup.php
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * Returns the entity with the provided id or null if there is no such
+	 * entity.
+	 *
+	 * @note Implementations of this method may or may not resolve redirects.
+	 * Code that needs control over redirect resolution should use an
+	 * EntityRevisionLookup instead.
+	 *
+	 * @since 0.3
```
JeroenDeDauw: .

src/Lookup/EntityRedirectLookup.php
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @licence GNU GPL v2+
+ * @author Daniel Kinzler
+ */
+interface EntityRedirectLookup {
+
+	/**
+	 * Returns the IDs that redirect to (are aliases of) the given target entity.
+	 *
+	 * @since 0.5
```
JeroenDeDauw: .

src/Lookup/EntityRedirectLookup.php
```diff
@@ -0,0 +1,42 @@
...
+	 * @param EntityId $targetId
+	 *
+	 * @return EntityId[]
+	 */
+	public function getRedirectIds( EntityId $targetId );
+
+	/**
+	 * Returns the redirect target associated with the given redirect ID.
+	 *
+	 * @since 0.5
```
JeroenDeDauw: .

src/Lookup/EntityLookup.php
```diff
@@ -0,0 +1,55 @@
...
+	 *
+	 * @param EntityId $entityId
+	 *
+	 * @throws OutOfBoundsException
```
JeroenDeDauw: Why and exception and the possibility of a returned null?? In case the exception is needed, having a more domain specific one, such as `EntityNotFoundException` would be nice. Did we not already have such an exception?

JeroenDeDauw:
Just realized that those parts where also unchanged... did not notice the oddness of that code when I looked at it earlier

addshore:
Well it doesn't look like the direct implementations throw these exceptions, they just come from deeper in the call stack

JeroenDeDauw:
That sounds like something that should not happen then. Exceptions should not bubble up like that.

JeroenDeDauw:
If we move over any such implementations we should probably fix that

PR #18, opened 2015-08-07
Add TypedValueFormatter & Exception

addshore:
This exception is new
The formatter is a CP from Wikibase lib

### DIFF
--- a/src/TypedValue/TypedValueFormatter.php
+++ b/src/TypedValue/TypedValueFormatter.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Wikibase\DataModel\Services\TypedValue;
+
+use DataValues\DataValue;
+
+/**
+ * Provides a string representation for a DataValue given its associated DataType.
+ *
+ * @since 0.4
+ *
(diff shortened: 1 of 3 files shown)


src/TypedValue/TypedValueFormatter.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Wikibase\DataModel\Services\TypedValue;
+
+use DataValues\DataValue;
+
+/**
+ * Provides a string representation for a DataValue given its associated DataType.
+ *
+ * @since 0.4
```
addshore: incorrect since tag

src/TypedValue/TypedValueFormatter.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Wikibase\DataModel\Services\TypedValue;
+
+use DataValues\DataValue;
```
addshore, line 5: need to fix composer.json to require this? :/
Didn't spot this one when looking through the list earlier

PR #20, opened 2015-08-07
Add PropertyLabelResolver

addshore:
This is a CP from WikibaseLib

### DIFF
--- a/src/PropertyLabelResolver.php
+++ b/src/PropertyLabelResolver.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Wikibase\DataModel\Services;
+
+use Wikibase\DataModel\Entity\EntityId;
+
+/**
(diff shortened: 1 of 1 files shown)


src/PropertyLabelResolver.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Wikibase\DataModel\Services;
```
JeroenDeDauw, line 3: Perhaps this should be in `\Term`? Or `\Lookup`?

addshore, line 3: I was thinking of Lookup, but this class does things differently to all of the Lookups we have (so doesn't quite fit)
It could go in Term though!

JeroenDeDauw, line 3: Moved in #26

PR #22, opened 2015-08-07
Add DiffOpFactory

addshore:
This is a CP from WikibaseLib
(with a minor alteration)

### DIFF
--- a/src/Diff/DiffOpFactory.php
+++ b/src/Diff/DiffOpFactory.php
@@ -1,10 +1,9 @@
 <?php
 
-namespace Wikibase;
+namespace Wikibase\DataModel\Services\Diff;
 
 use Diff\DiffOp\DiffOp;
 use Diff\DiffOpFactory as DiffDiffOpFactory;
-use Wikibase\DataModel\Services\Diff\EntityDiff;
 
 /**
  * Class for changes that can be represented as a Diff.
--- a/src/Diff/DiffOpFactory.php
+++ b/src/Diff/DiffOpFactory.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Wikibase;
+
+use Diff\DiffOp\DiffOp;
+use Diff\DiffOpFactory as DiffDiffOpFactory;
+use Wikibase\DataModel\Services\Diff\EntityDiff;
+
+/**
+ * Class for changes that can be represented as a Diff.
+ *
+ * @since 1.1
+ *
+ * @licence GNU GPL v2+
+ * @author Daniel Kinzler
+ */
+class DiffOpFactory extends DiffDiffOpFactory {
+
+	public function newFromArray( array $diffOp ) {
+		$this->assertHasKey( 'type', $diffOp );
+
+		// see EntityDiff::getType() and ItemDiff::getType()
+		if ( preg_match( '!^diff/(.*)$!', $diffOp['type'], $matches ) ) {
+			$itemType = $matches[1];
+			$this->assertHasKey( 'operations', $diffOp );
+
+			$operations = $this->createOperations( $diffOp['operations'] );
+			$diff = EntityDiff::newForType( $itemType, $operations );
+
+			return $diff;
+		}
+
+		return parent::newFromArray( $diffOp );
+	}
+
+	/**
+	 * Converts a list of diff operations represented by arrays into a list of
+	 * DiffOp objects.
+	 *
+	 * @todo: pull this up into DiffOpFactory
+	 *
+	 * @param array $data the input data
+	 * @return DiffOp[] The diff ops
+	 */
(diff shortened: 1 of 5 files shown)


src/Diff/DiffOpFactory.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Wikibase;
+
+use Diff\DiffOp\DiffOp;
+use Diff\DiffOpFactory as DiffDiffOpFactory;
```
JeroenDeDauw: :)

src/Diff/DiffOpFactory.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Wikibase;
```
JeroenDeDauw: NS. The code is broken as it, since it won't get autoloaded. Lack of test is bad...

JeroenDeDauw:
What is the minor alteration?

src/Diff/DiffOpFactory.php
addshore: This was the minor alteration ;) So that this class doesn't have to be called WikibaseDiffOpFactory

addshore:
See the inline comment, I renamed the class basically ;)

JeroenDeDauw:
As far as I can tell this test was copied from elsewhere and just tests the behaviour of the superclass (which is concrete and already tested). I'd rather see a test for the behaviour the class itself adds.

src/Diff/DiffOpFactory.php
```diff
@@ -0,0 +1,59 @@
...
+namespace Wikibase\DataModel\Services\Diff;
+
+use Diff\DiffOp\DiffOp;
+use Diff\DiffOpFactory as DiffDiffOpFactory;
+use InvalidArgumentException;
+
+/**
+ * Class for changes that can be represented as a Diff.
+ *
+ * @since 1.1
```
thiemowmde: 1.2?

addshore: Yep, this would now be going in for 1.2

JeroenDeDauw: Updated

thiemowmde:
-1 for having two classes with the exact same name.

Not sure what's wrong with the test. Without this the class would have zero coverage.

JeroenDeDauw:
We are not adding a new class or renaming one. This is a move from Wikibase Lib to this component. They already have the same name.

My concern about the test has already been addressed

thiemowmde:
Sure this class got renamed, @addshore talked about this above. It was called `WikibaseDiffOpFactory` in Lib. It's basically `class \Wikibase\DataModel\Services\Diff\DiffOpFactory extends \Diff\DiffOpFactory` now and I disagree with two classes, one extending the other, having the exact same name. The extended one should make it's purpose clear with a more specific name.

JeroenDeDauw:
#fail >_> addshore even wrote that in the comment above.

addshore:
The purpose is clear through the namespace of the class.
I can't see any case where these 2 classes are going to be used side by side.
The Wikibase prefix to me seems quite pointless, but if others want it I'll add it back

thiemowmde:
~~I just realized that this class being a subclass of `\Diff\DiffOpFactory` is completely pointless. It does not use **any** code from it's parent except for the trivial assertion. Should the parent class be abstract (and renamed accordingly)?~~

~~Even simpler, just ditch the `extends` and ditch the whole base class. It's completely unused in our code base. Then you are free to name the remaining class after the former base class.~~

thiemowmde:
Even simpler, just ditch the `extends` and ditch the whole base class. It's completely unused in out code base. Then you are free to name the remaining class after the former base class.

JeroenDeDauw:
I guess you did not see the `return parent::newFromArray( $diffOp );`?

Removing the baselcass from Diff is somewhat odd as it mirrors the `toArray` behaviour of the operations defined in Diff.

thiemowmde:
Yes, I missed the `parent` call. Ok, then this class should really have a more specific name that reflects the purpose and difference to the base class.

addshore:
Well, if anyone has any suggestion I will happily amend this....
My thoughts on the name can already be found @ https://github.com/wmde/WikibaseDataModelServices/pull/22#issuecomment-130344511

JeroenDeDauw:
I'm fine with this as it is if these things will not be used together, which indeed seems likely. Or with renaming it. +2 from my side, feel free to merge. A follow up can be submitted if someone wants to change it.

thiemowmde:
+2 if #45 is merged first.

addshore:
BAH, needs a rebase with that merged ontop apparently....

thiemowmde:
Rebased. +2 from my side now that #45 is merged in here, so feel free to merge.

PR #27, opened 2015-08-09
Move over EntityRetrievingTermLookup

JeroenDeDauw:
This class was copied from Wikibase Lib.

Some alternations where made. fetchFingerprint now catches Exception
rather than two more specific ones and has a TODO (#28) there as it should
not be catching an exception as per EntityLookup interface contract.

The test was also modified to use a new trivial EntityLookup implementation,
rather than the 900 line and rather complex MockRepository class.

### DIFF
--- a/tests/fixtures/InMemoryEntityLookup.php
+++ b/tests/fixtures/InMemoryEntityLookup.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Fixtures;
+
+use Wikibase\DataModel\Entity\EntityDocument;
+use Wikibase\DataModel\Entity\EntityId;
+use Wikibase\DataModel\Services\Lookup\EntityLookup;
+
+class InMemoryEntityLookup implements EntityLookup {
+
+	private $entities = array();
+
+	public function addEntity( EntityDocument $entity ) {
+		if ( $entity->getId() === null ) {
+			throw new \InvalidArgumentException( 'The entity needs to have an ID' );
+		}
+
+		$this->entities[$entity->getId()->getSerialization()] = $entity;
+	}
+
+	public function getEntity( EntityId $entityId ) {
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Please also see #29 (answering that does not block merging this)

tests/fixtures/InMemoryEntityLookup.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Fixtures;
+
+use Wikibase\DataModel\Entity\EntityDocument;
+use Wikibase\DataModel\Entity\EntityId;
+use Wikibase\DataModel\Services\Lookup\EntityLookup;
+
+class InMemoryEntityLookup implements EntityLookup {
```
addshore, line 9: @since? :)

tests/unit/Lookup/EntityRetrievingTermLookupTest.php
```diff
@@ -0,0 +1,154 @@
...
+use Wikibase\DataModel\Services\Lookup\EntityRetrievingTermLookup;
+
+/**
+ * @covers Wikibase\DataModel\Services\Lookup\EntityRetrievingTermLookup
+ *
+ * @licence GNU GPL v2+
+ * @author Katie Filbert < aude.wiki@gmail.com >
+ * @author Daniel Kinzler
+ */
+class EntityRetrievingTermLookupTest extends \PHPUnit_Framework_TestCase {
```
addshore: @since? :)

tests/fixtures/InMemoryEntityLookup.php
JeroenDeDauw, line 9: This class is not package public

addshore, line 9: _facepalm_ indeed....

PR #31, opened 2015-08-09
Remove @group tags left in some of the moved tests

JeroenDeDauw:


Benestar:
Since when do we avoid having `@group` tags in tests? I think some users might rely on the `--group` feature PHPUnit provides.

JeroenDeDauw:
> Since when do we avoid having @group tags in tests?

About two years ago I guess. We discussed that internally on the list and in some meetings. Perhaps people are still doing this in some places in WB.git (because there we need to use the stupid MW test runner), though in our components we no longer have these.

addshore:
+1
I don't use groups, even through the mediawiki test runner...

mariushoch:
I'm not sure the MediaWiki test runner will pick these up, without `@group Wikibase`, which is especially bad for the build. Can you look at this @JanZerebecki?

JeroenDeDauw:
@mariushoch these classes are not registered with the MediaWiki test runner to begin with. If you want to investigate if we should change that, fine, though out of scope here. We've been not using @group since forever for our components.

Benestar:
So do I understand it right that those tags should also be removed from WikibaseDataModel?

JeroenDeDauw:
I just checked and am surprised by the number of tags left there. Still, they are by far the exception rather than the rule.

PR #32, opened 2015-08-09
Add EntityRetrievingDataTypeLookup, moved from Wikibase Lib

JeroenDeDauw:
This depends on #27, so that is also in the commit list here now until #27 gets merged.

### DIFF
--- a/tests/fixtures/InMemoryEntityLookup.php
+++ b/tests/fixtures/InMemoryEntityLookup.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Fixtures;
+
+use Wikibase\DataModel\Entity\EntityDocument;
+use Wikibase\DataModel\Entity\EntityId;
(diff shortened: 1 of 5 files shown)


tests/fixtures/InMemoryEntityLookup.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Fixtures;
+
+use Wikibase\DataModel\Entity\EntityDocument;
+use Wikibase\DataModel\Entity\EntityId;
+use Wikibase\DataModel\Services\Lookup\EntityLookup;
+
+class InMemoryEntityLookup implements EntityLookup {
```
addshore, line 9: @since?

JeroenDeDauw, line 9: This class is not package public

addshore, line 9: _facepalm again_.....
This is why CR should not be the first thing you do in the day!

PR #38, opened 2015-08-11
Improve PHPUnit compatibility

thiemowmde:
I would love to use these methods, but they do not exist in PHPUnit version we still use. The shorter code is not worth the reduced compatibility. That's why no other test in our whole code base uses these new methods. This are the only 2 code lines.

### DIFF
--- a/tests/unit/Lookup/LanguageLabelDescriptionLookupTest.php
+++ b/tests/unit/Lookup/LanguageLabelDescriptionLookupTest.php
@@ -20,7 +20,7 @@ class LanguageLabelDescriptionLookupTest extends \PHPUnit_Framework_TestCase {
 		$termLookup->expects( $this->once() )
 			->method( 'getLabel' )
 			->with( $this->equalTo( new ItemId( 'Q42' ) ), $this->equalTo( 'language_code' ) )
-			->willReturn( 'term_text' );
+			->will( $this->returnValue( 'term_text' ) );
 
 		$lookup = new LanguageLabelDescriptionLookup( $termLookup, 'language_code' );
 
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
> PHPUnit version we still use

Who is "we"? And why are you still using 3.x after 4.x has been released for so long already?

thiemowmde:
I will not respond to a post tearing my arguments apart. If you disagree with one, fine. There are others.

JeroenDeDauw:
To be clear, I do not object to this change and would probably merge it if my questions where answered. The reason I'm asking them is because I do not agree at present with us needing to keep compatibility with PHPUnit 3.x everywhere, and am curious why someone would want to do that. Certainly we are not doing what we should be doing if we'd want to commit to retaining such compatibility.

thiemowmde:
I honestly do not understand the last sentence.

I do not find your questions honest and open:
- "Why are you" - as if nobody else would ever try to run our tests with PHPUnit 3.
- "still", "has been released for so long already" - as if I'm to lazy to update and that would be my only motivation.
- "using" - as if I do not have an alternative and are not running previous versions on purpose.

By asking who "we" is you are tearing my argument apart, ripping words out of context. It does not matter at all who "we" is. Pick what you want. "You", "I", "other developers", "people", it does not matter and does not change the argument. All tests run fine with PHPUnit 3. Whoever tries to run them with version 3 will find that these two lines (in the **whole** Wikibase code base) are the only reason why it will not run. And will ask why we decided to dish such an easy to keep compatibility for what's obviously a negligible benefit.

I would really, really love to not have such discussions. But if patches repeatedly got stuck for literally nothing, no matter if I respond to leading questions or refuse to do so, what else could I do? I honestly have no idea.

JeroenDeDauw:
I'm asking the questions I'm asking because they are relevant to me. You go to a lot of effort to state that they do not matter to you and that they amount to nothing. That you are getting upset when I question the need for supporting an old version of some software or something similar is your own doing. It's reasonable to ask such questions and to question the premises on which you build your arguments. That is part of code review.

thiemowmde:
So you really don't understand what the problem is with the way you are asking questions?

PR #39, opened 2015-08-11
Remove --coverage-php option

thiemowmde:
The difference (on my machine) is ~190 ms versus ~1.5 seconds. That's sooo much slower that it's not worth the benefit, in my opinion. It would be cool to have this coverage test run once a day or something like this, but it should not slow down CI.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
 	"scripts": {
 		"test": [
 			"composer validate --no-interaction",
-			"phpunit --coverage-php /dev/null"
+			"phpunit"
 		],
 		"cs": [
 			"composer phpcs",


JeroenDeDauw:
This check costs one or two second yes. How is that not worth it? We've already lost more time by the check not being there than I expect we're ever going to lose by having this.

thiemowmde:
Answered at https://github.com/DataValues/Common/pull/33#issuecomment-129782318.

addshore:
First good use of having this https://travis-ci.org/wmde/WikibaseDataModelServices/jobs/75920078

Interestingly for php7 it padded.... https://travis-ci.org/wmde/WikibaseDataModelServices/builds/75920076

JeroenDeDauw:
If I'm not mistaken that's actually the third time it caught something

PR #40, opened 2015-08-12
Add RedirectResolvingEntityLookup, moved from Wikibase Lib

JeroenDeDauw:
Copy-paste, not changed to the code appart from the NS difference

### DIFF
--- a/src/Lookup/RedirectResolvingEntityLookup.php
+++ b/src/Lookup/RedirectResolvingEntityLookup.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Lookup;
+
+use Wikibase\DataModel\Entity\EntityDocument;
+use Wikibase\DataModel\Entity\EntityId;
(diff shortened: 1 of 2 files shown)


thiemowmde:
+1, looks good.

addshore:
Per the discussion we just had on IRC this is going to end up throwing something

JeroenDeDauw:
Well yeah sure. I did not change anything here compared to WB.git though. So I think its easiest to address this if we merge this class and then make a PR that adds an exception and documents it in all the docblocks.

thiemowmde:
> I did not change anything here compared to WB.git though.

What do you mean? This copy misses the two `@throws StorageException` tags in question. They are not just comments but part of the contract of the class - something you taught me.

JeroenDeDauw:
If I removed any such comments then I forgot they where there in the first place when I wrote that comment, I certainly can't remember them now.

addshore:
These will be added back in with or after https://github.com/wmde/WikibaseDataModelServices/pull/42 before the next release

PR #41, opened 2015-08-12
Use self in EntityDiff

JeroenDeDauw:


thiemowmde:
+2, obviously, when tests pass.

PR #42, opened 2015-08-14
Add EntityNotFoundException & use in EntityLookup

addshore:
This breaks the last contract of this interface.

getEntity should no longer return null but should
throw an exception instead.


JeroenDeDauw:
Rel notes need updating

addshore:
yup

PR #44, opened 2015-08-17
Cleanup throws and null returns in all Lookups RFC

addshore:
This is going to start of as an RFC so that I can get peoples general thoughts on this and ideas / comments.
Right now the lookups do odd stuff in odd ways etc. and this is trying to fix all of that....

This tries to fix all of the lookups to do things the same way.

When trying to lookup a value that does not exist then return null.
In the case we were looking for multiple values / an array then
return an empty array.

If something when wrong during the lookup then throw an exception.
EntityIdLookupException has been created for this, which could also
be called a ByEntityIdLookupException.....


JeroenDeDauw:
I agree with the null for not found and exception for exceptional behavior.

The approach of having only a single exception like that is interesting. While it can work, I do not think this is a good idea for cases such as `TermLookup` and `PropertyDataTypeLookup`. The methods in the former have additional info and the one in the later always has a `PropertyId`. And if you already have multiple exceptions due to that anyway, then why not go for dedicated ones per lookup as well, so you have domain specific exceptions everywhere?

I can't remember seeing a place where exceptions where created based on signature rather than domain - if you know one I'd be interested in seeing it.

addshore:
Removed the RFC bit

src/Lookup/PropertyLookup.php
```diff
@@ -18,8 +18,8 @@
 	 *
 	 * @param PropertyId $propertyId
 	 *
-	 * @return Property
-	 * @throws PropertyNotFoundException
+	 * @return Property|null
+	 * @throws EntityLookupException
```
JeroenDeDauw: How about `PropertyLookupException`? That then also knows it has a `PropertyId`

src/Lookup/PropertyDataTypeLookupException.php
```diff
@@ -0,0 +1,34 @@
...
+ * @since 1.2
+ *
+ * @licence GNU GPL v2+
+ * @author Adam Shorland
+ */
+class PropertyDataTypeLookupException extends \RuntimeException {
+
+	private $entityId;
+
+	public function __construct( EntityId $entityId, $message = null, \Exception $previous = null ) {
```
JeroenDeDauw: `PropertyId`

src/Lookup/TermLookupException.php
```diff
@@ -0,0 +1,41 @@
...
+ * @licence GNU GPL v2+
+ * @author Adam Shorland
+ */
+class TermLookupException extends \RuntimeException {
+
+	private $entityId;
+
+	public function __construct(
+		EntityId $entityId,
+		array $languageCodes,
```
JeroenDeDauw, line 19: The array bit is a bit odd, since some of the methods just take one. I don't think this is a real issue, just saying it's a bit odd.

JeroenDeDauw:
These are breaking changes, so those @since should be 2.0

JeroenDeDauw:
:+1:

addshore:
!Merg!

JeroenDeDauw:
Sure. I want another pair of eyes to double check before release though. Getting something wrong in a breaking release and being forced to do another breaking one would be fail

src/Lookup/EntityRetrievingTermLookup.php
```diff
@@ -38,11 +39,20 @@ public function __construct( EntityLookup $entityLookup ) {
...
 	 */
 	public function getLabel( EntityId $entityId, $languageCode ) {
-		$labels = $this->getFingerprint( $entityId )->getLabels();
-		return $labels->getByLanguage( $languageCode )->getText();
+		$fingerprint = $this->getFingerprint( $entityId, array( $languageCode ) );
+
+		/** @var Fingerprint $fingerprint */
+		$labels = $fingerprint->getLabels();
+
+		try{
```
Benestar, line 51: missing space :hear_no_evil:

addshore, line 51: Oh nooooo, the world is going to end!

addshore, line 51: Fixed in https://github.com/wmde/WikibaseDataModelServices/commit/180b24fc3e2be759d61a9098d76f23c7d94a8c45

JeroenDeDauw, line 51: Our CS checks did not catch this?

addshore, line 51: apparently not

PR #46, opened 2015-08-18
Revert to use non-deprecated Item::copy and Property::copy

thiemowmde:
This reverts parts of https://gerrit.wikimedia.org/r/#/c/222978/. See https://github.com/wmde/WikibaseDataModel/pull/531 for arguments. In short: Yes, `Entity` is deprecated and `Entity::copy` should not be called. But `Item::copy` and `Property::copy` are **not** deprecated.


JeroenDeDauw:
Good catch

PR #54, opened 2015-08-26
Remove false from EntityRedirectLookup

thiemowmde:
This should block the 2.0.0 release because (I think) I found an _other_ mistake in addition to the one fixed in #57.


addshore:
_facepalm_ Why is this all in one commit / PR?
If the changes the readme and the code changes were separate I could have merged one of them!!!

JeroenDeDauw:
Nice catch. I've extracted the production code change, added tests and submitted that as #57

PR #60, opened 2015-09-01
Add SiteLinkLookup from Wikibase Lib

Benestar:


JeroenDeDauw:
~~`$pageTitle` is a MediaWiki concept. Should not be in this component~~

Benestar:
That's just the naming of the variable. It's obvious that the method is just another version of `getItemIdForSiteLink`. We can however adjust the naming of the variables. That doesn't really change the interface though.

addshore:
We could just kill the getItemIdForLink in the interface?

Benestar:
Even better idea @addshore #killallthethings

addshore:
hah, I think I am getting a name for myself when it comes to killing things...
Either kill it or rename the params to $siteId and $pageName as in https://github.com/wmde/WikibaseDataModel/blob/master/src/SiteLink.php#L47

Benestar:
We should still have the method in there. There are cases where we deal with title objects and site ids independently and we perhaps don't want to construct a SiteLink object for no reason.

JeroenDeDauw:
I'm looking through the git history for this interface out of curiosity. Originally it just had

``` php
public function getConflictsForItem( Item $item );
public function getItemIdForLink( $globalSiteId, $pageTitle );
```

First degradation happened in https://gerrit.wikimedia.org/r/#/c/32172/4/lib/includes/store/SiteLinkLookup.php by me :) After that it grew more and some stuff got removed

JeroenDeDauw:
I'm thinking this interface is probably not well segregated, and should be split based on the users of the methods. While for the initial move from WB.git we figured to just move and improve later, I suspect it's easier now to first improve and then move. This avoids forcing breaking releases of this component and should be easy enough to manage, as we are not moving dozens of things and in the process of introducing new component.

Does someone know if any of the interfaces methods are frequently used together, or used together at all?

src/Lookup/SiteLinkLookup.php
```diff
@@ -0,0 +1,72 @@
...
+	 * @return ItemId|null
+	 */
+	public function getItemIdForLink( $globalSiteId, $pageTitle );
+
+	/**
+	 * Returns the links that match the provided conditions.
+	 * The links are returned as arrays with the following elements in specified order:
+	 * - string siteId
+	 * - string pageName
+	 * - int itemId Numeric (unprefixed) item id
```
thiemowmde, line 36: This is a weird interface. An array with three numerically indexed elements? I do not think such an interface should be in this generic services component. This sounds like a concept that's specific to Wikibase.git and should stay there.

src/Lookup/SiteLinkLookup.php
```diff
@@ -0,0 +1,72 @@
...
+interface SiteLinkLookup {
+
+	/**
+	 * Returns the id of the item that is equivalent to the
+	 * provided page, or null if there is none.
+	 *
+	 * @since 2.0
+	 *
+	 * @param string $globalSiteId
+	 * @param string $pageTitle
```
thiemowmde, line 25: Similar problem here. Why does this use two strings instead of a SiteLink object?

addshore, line 25: I am guessing there are cases when we already have the strings and no object.
Slightly quicker to then just pass the strings in.
We also might not have the badges in such cases and thus creating an object with empty / possibly incorrect badges is a bit odd.

JeroenDeDauw:

PR #61, opened 2015-09-02
Add ItemAndPropertyDiffer

thiemowmde:
This is a resubmission of https://github.com/wmde/WikibaseDataModel/pull/354.

This patch introduces a generic differ that allows comparing items and properties. Note that both do have statements now. The only remaining difference is that properties don't have site links.

### DIFF
--- a/phpmd.xml
+++ b/phpmd.xml
@@ -23,7 +23,14 @@
         </properties>
     </rule>
 
-    <rule ref="rulesets/design.xml" />
+    <rule ref="rulesets/design.xml">
+		<exclude name="CouplingBetweenObjects" />
+    </rule>
+	<rule ref="rulesets/design.xml/CouplingBetweenObjects">
+		<properties>
+			<property name="minimum" value="15" />
+		</properties>
+	</rule>
 
     <rule ref="rulesets/naming.xml">
         <exclude name="ShortVariable" />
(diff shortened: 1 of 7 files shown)


JeroenDeDauw:
> The only remaining difference is that properties don't have site links.

And that items don't have a data type.

Benestar:
I wonder where we need such a generic differ. Why would we want to compare items and properties? I oppose to introduce a differ that has the word "generic" in it as it is not generic but only works for items and properties. Furthermore, without a usecase for diffing items with properties I don't understand where htis would be useful.

thiemowmde:
> only works for items and properties.

This is not true. It **is** a generic differ that works in all cases when a more specific differ can not be found.

> without a usecase for diffing items with properties

Hold your breath, here comes the use case: I want to compare items with properties.

I can compare lists of statements across different entities if they share the same namespace: https://www.wikidata.org/w/index.php?title=Q7560&oldid=246252141&diff=245706731. Why am I not allowed to do the same across namespaces: https://www.wikidata.org/w/index.php?title=Q7560&oldid=246252141&diff=238881024? It's basically the single `if ... throw` I'm removing in this patch that blocks this. Unblocking this use case is so easy that I really wonder why it was blocked in the first place.

JeroenDeDauw:

PR #62, opened 2015-09-02
2.0.0 release (perhaps really this time ;p)

JeroenDeDauw:


.travis.yml
```diff
@@ -10,7 +10,7 @@ matrix:
       php: 5.3
     - env: DM=~4.2
       php: 5.4
-    - env: DM=~4.2
+    - env: DM=~4.3@dev
```
Benestar, line 13: we already test against DM@dev below with the same configuration. The dev version should always allow failures as it shouldn't block pull requests on this component.

Benestar:
<s>-1 to the travis config change. A broken and unreleased DM master shouldn't make the tests fail on an unrelated component.</s>

Benestar:
per discussion on irc

PR #67, opened 2015-09-15
Address exception inconsistencies in RedirectResolvingEntityLookup

JeroenDeDauw:


src/Lookup/UnresolvedEntityRedirectException.php
```diff
@@ -22,10 +21,11 @@ class UnresolvedRedirectException extends RuntimeException {
 	private $redirectTargetId;
 
 	/**
+	 * @param EntityId $entityId
 	 * @param EntityId $redirectTargetId The ID of the target Entity of the redirect
 	 */
-	public function __construct( EntityId $redirectTargetId ) {
-		parent::__construct( "Unresolved redirect to " . $redirectTargetId->getSerialization() );
+	public function __construct( EntityId $entityId, EntityId $redirectTargetId ) {
+		parent::__construct( $entityId, "Unresolved redirect to " . $redirectTargetId->getSerialization() );
```
brightbyte, line 28: Wikibase still needs the revision ID and Timestamp here. Or shall we use a subclass?

Benestar, line 28: How is this related to the proposed change? I think the revision ID and timestamp weren't there before in this component.

JeroenDeDauw, line 28: Subclassing is one thing that can be done. Keep in mind that not all code in Wikibase.git needs this extra info, just some of it. Loose from these concepts not belonging in this component, they should not have been put into UnresolvedRedirectException to begin with.

brightbyte, line 28: @bene: they used to be in this exception, and wikibase needs them there. they got removed when the exception was moved to the component.

Benestar, line 28: Ok, but this shouldn't block this patch in particular then but can be discussed elsewhere, right? As far as I can see this only refactors some classes and renames an interface.

JeroenDeDauw, line 28: What I did in this PR actually allow us to solve the issue referred here.

PR #69, opened 2015-09-17
Add optional message and previous parameters to UnresolvedEntityRedir…

JeroenDeDauw:
…ectException

### DIFF
--- a/src/Lookup/UnresolvedEntityRedirectException.php
+++ b/src/Lookup/UnresolvedEntityRedirectException.php
@@ -2,6 +2,7 @@
 
 namespace Wikibase\DataModel\Services\Lookup;
 
+use Exception;
 use Wikibase\DataModel\Entity\EntityId;
 
 /**
@@ -23,9 +24,15 @@ class UnresolvedEntityRedirectException extends EntityLookupException {
 	/**
 	 * @param EntityId $entityId
 	 * @param EntityId $redirectTargetId The ID of the target Entity of the redirect
+	 * @param string|null $message
+	 * @param Exception|null $previous
 	 */
-	public function __construct( EntityId $entityId, EntityId $redirectTargetId ) {
-		parent::__construct( $entityId, 'Unresolved redirect to ' . $redirectTargetId->getSerialization() );
+	public function __construct( EntityId $entityId, EntityId $redirectTargetId, $message = null, Exception $previous = null ) {
+		parent::__construct(
+			$entityId,
+			$message !== null ? $message : 'Unresolved redirect to ' . $redirectTargetId->getSerialization(),
+			$previous
+		);
 
 		$this->redirectTargetId = $redirectTargetId;
 	}


src/Lookup/UnresolvedEntityRedirectException.php
```diff
@@ -23,9 +24,15 @@ class UnresolvedEntityRedirectException extends EntityLookupException {
...
 	 * @param EntityId $redirectTargetId The ID of the target Entity of the redirect
+	 * @param string|null $message
+	 * @param Exception|null $previous
 	 */
-	public function __construct( EntityId $entityId, EntityId $redirectTargetId ) {
-		parent::__construct( $entityId, "Unresolved redirect to " . $redirectTargetId->getSerialization() );
+	public function __construct( EntityId $entityId, EntityId $redirectTargetId, $message = null, Exception $previous = null ) {
+		parent::__construct(
+			$entityId,
+			$message !== null ? $message : "Unresolved redirect to " . $redirectTargetId->getSerialization(),
```
thiemowmde: ~~So the two parameters `$redirectTargetId` and `$message` are now exclusive? I find this odd. Or to ask this via a different question: What's the use of a custom message?~~

+1 for `$previous`.

src/Lookup/UnresolvedEntityRedirectException.php
```diff
@@ -23,9 +24,15 @@ class UnresolvedEntityRedirectException extends EntityLookupException {
 	/**
 	 * @param EntityId $entityId
```
thiemowmde, line 25: Does this now need a `@since 3.1`?

JeroenDeDauw, line 25: I'll add the tags together with release notes update after merge.

thiemowmde:
~~As said inline, I would vote -1 because I find the `$message` parameter confusing and probably useless.~~

src/Lookup/UnresolvedEntityRedirectException.php
JeroenDeDauw: > So the two parameters $redirectTargetId and $message are now exclusive?

No. The message that will be constructed when none is provided uses $redirectTargetId. That is as far as the relation between those two parameters goes.

> Or to ask this via a different question: What's the use of a custom message?

To allow the things throwing the extension to provide a more specific error message. Same reason as why Exception itself has a message parameter.

src/Lookup/UnresolvedEntityRedirectException.php
```diff
@@ -23,9 +24,15 @@ class UnresolvedEntityRedirectException extends EntityLookupException {
...
 	 * @param EntityId $redirectTargetId The ID of the target Entity of the redirect
+	 * @param string|null $message
+	 * @param Exception|null $previous
 	 */
-	public function __construct( EntityId $entityId, EntityId $redirectTargetId ) {
-		parent::__construct( $entityId, 'Unresolved redirect to ' . $redirectTargetId->getSerialization() );
+	public function __construct( EntityId $entityId, EntityId $redirectTargetId, $message = null, Exception $previous = null ) {
```
thiemowmde, line 30: ~~So when a custom `$message` is provided, the `$redirectTargetId` is ignored. So `$message` does have the higher precedence, but comes later in the list of parameters and is optional. `$redirectTargetId` on the other hand is mandatory, even if a custom `$message` is provided and `$redirectTargetId` is ignored. Still a -1 for this.~~

JeroenDeDauw, line 30: Did you see the line that does this?

``` php
$this->redirectTargetId = $redirectTargetId;
```

thiemowmde, line 30: D'oh!

src/Lookup/UnresolvedEntityRedirectException.php
```diff
@@ -23,9 +24,15 @@ class UnresolvedEntityRedirectException extends EntityLookupException {
...
+		parent::__construct(
+			$entityId,
+			$message !== null ? $message : 'Unresolved redirect to ' . $redirectTargetId->getSerialization(),
```
thiemowmde, line 33: I like the `$message ?: ...` shortcut in such cases. Yes, I know, the `'0'` edge case. But seriously, who uses "0" as a message?

PR #71, opened 2015-10-01
Add exception support to InMemoryEntityLookup and add tests

JeroenDeDauw:
> i think some of the problem is that while we have InMemoryEntityLookup
> (and think it would be nice to use in tests), it does not cover exception
> handling stuff
> 
> it would be nice to have a consistent thing to use in tests instead of
> making a mock callback in places that tries to mimic EntityLookup

-- @filbertkm

### DIFF
--- a/src/Lookup/InMemoryEntityLookup.php
+++ b/src/Lookup/InMemoryEntityLookup.php
@@ -37,6 +37,7 @@ class InMemoryEntityLookup implements EntityLookup {
 	}
 
 	/**
+	 * Registers an exception that will be thrown when a entity with the id in the exception is requested.
 	 * If an exception with the same EntityId was already present it will be replaced by the new one.
 	 *
 	 * @since 3.1
--- a/src/Lookup/InMemoryEntityLookup.php
+++ b/src/Lookup/InMemoryEntityLookup.php
@@ -8,6 +8,9 @@ use Wikibase\DataModel\Entity\EntityId;
 
 /**
  * EntityLookup that uses an in memory array to retrieve the requested information.
+ * One can also specify exceptions that should be thrown when an entity with their
+ * associated ID is requested.
+ *
  * This class can be used as a fake in tests.
  *
  * @since 2.0
@@ -18,6 +21,7 @@ use Wikibase\DataModel\Entity\EntityId;
 class InMemoryEntityLookup implements EntityLookup {
 
 	private $entities = array();
+	private $exceptions = array();
 
 	/**
 	 * @param EntityDocument $entity
@@ -33,6 +37,17 @@ class InMemoryEntityLookup implements EntityLookup {
 	}
 
 	/**
+	 * If an exception with the same EntityId was already present it will be replaced by the new one.
+	 *
+	 * @since 3.1
+	 *
+	 * @param EntityLookupException $exception
+	 */
+	public function addException( EntityLookupException $exception ) {
+		$this->exceptions[$exception->getEntityId()->getSerialization()] = $exception;
+	}
+
+	/**
 	 * @see EntityLookup::getEntity
 	 *
 	 * @param EntityId $entityId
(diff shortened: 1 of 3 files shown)


thiemowmde:
Code looks fine, but personally I find the names "fake" and "evil" irritating. How is this a "fake"? What does it mean when something is "evil"?

JeroenDeDauw:
I removed the evil part. A fake is a type of test double, see https://en.wikipedia.org/wiki/Mock_object#Mocks.2C_Fakes_and_Stubs

Benestar:
I think a more specific name would be good. `Fake` is ok but not really descriptive. If someone can come up with a better name than lets use that. Otherwise, we can also use the `Fake` name.

JeroenDeDauw:
Not sure what made me create a new class. It can be in the existing one without causing any harm, as exceptions are part of the interface. Changed now.

src/Lookup/InMemoryEntityLookup.php
```diff
@@ -18,6 +21,7 @@
 class InMemoryEntityLookup implements EntityLookup {
 
 	private $entities = array();
+	private $exceptions = array();
```
thiemowmde, line 24: Me missing type docs. :-(

src/Lookup/InMemoryEntityLookup.php
```diff
@@ -33,6 +37,17 @@ public function addEntity( EntityDocument $entity ) {
 	}
 
 	/**
+	 * If an exception with the same EntityId was already present it will be replaced by the new one.
```
thiemowmde, line 41: This is not actually a description of the purpose of these exceptions.

thiemowmde:
I love it! +2 for the code, ~~but I'm missing a bit of documentation. See inline comments~~. Don't forget to update the release notes.

src/Lookup/InMemoryEntityLookup.php
JeroenDeDauw, line 41: Indeed. It points out behaviour of this method that would otherwise require the user to look in its implementation to find out. Do you think something should be changed here?

src/Lookup/InMemoryEntityLookup.php
JeroenDeDauw, line 24: The types are documented in the public places, and can be inferred here by any non tard tool.

src/Lookup/InMemoryEntityLookup.php
thiemowmde, line 41: A basic explanation of what the function does. Currently this only describes a side effect. E.g. "Collects exceptions that should be thrown when a entity with the id in the exception is requested."

src/Lookup/InMemoryEntityLookup.php
thiemowmde, line 24: I know what you mean, but I still disagree. Especially here where the type `EntityLookupException[]` is nowhere else in this file. What this array contains only becomes clear when you search the code for all places that add stuff to this array. You need to read the **whole** file, not just a constructor or something.

Not critical, so feel free to merge.

JeroenDeDauw, line 24: If you start by looking at private implementation details rather than the public interface, then yeah. No need to read the whole file - just look at the value writes, which is one location, which has the type on the line above it.

By all means feel free to add the type hints though. Somewhat cluttery to me, though not an issue.

src/Lookup/InMemoryEntityLookup.php
```diff
@@ -53,10 +71,19 @@ public function getEntity( EntityId $entityId ) {
...
+	 * @throws EntityLookupException
 	 * @return bool
 	 */
 	public function hasEntity( EntityId $entityId ) {
+		$this->throwExceptionIfNeeded( $entityId );
+
 		return array_key_exists( $entityId->getSerialization(), $this->entities );
 	}
 
+	private function throwExceptionIfNeeded( EntityId $entityId ) {
```
filbertkm, line 83: "IfNeeded" is slightly odd but don't have a better suggestion right now :/

filbertkm:
I think this will be generally helpful and its much nicer to support this functionality within the InMemoryEntityLookup as this change does now. (vs in a separate class)

PR #75, opened 2015-10-09
Introduce PropertyDataTypeMatcher

filbertkm:
This wraps around PropertyDataTypeLookup to check if a Property has a given data type, with in process caching for performance reasons.

### DIFF
--- a/src/Lookup/InProcessCachingDataTypeLookup.php
+++ b/src/Lookup/InProcessCachingDataTypeLookup.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Lookup;
+
+use Wikibase\DataModel\Entity\PropertyId;
+
+/**
+ * PropertyDataTypeLookup that provides in-process caching.
+ *
+ * @since 3.1
+ *
+ * @licence GNU GPL v2+
+ * @author Katie Filbert < aude.wiki@gmail.com >
+ */
+class InProcessCachingDataTypeLookup implements PropertyDataTypeLookup {
+
+	/**
+	 * @var string[] Indexed by serialized PropertyId
+	 */
+	private $propertyIds = array();
+
+	/**
+	 * @param PropertyDataTypeLookup $propertyDataTypeLookup
+	 */
+	public function __construct( PropertyDataTypeLookup $propertyDataTypeLookup ) {
+		$this->propertyDataTypeLookup = $propertyDataTypeLookup;
+	}
+
+	/**
+	 * @param PropertyId $propertyId
+	 *
+	 * @return string
+	 * @throws PropertyDataTypeLookupException
+	 */
+	public function getDataTypeIdForProperty( PropertyId $propertyId ) {
+		$serializedId = $propertyId->getSerialization();
+
+		if ( !array_key_exists( $serializedId, $this->propertyIds ) ) {
+			$dataTypeId = $this->propertyDataTypeLookup->getDataTypeIdFOrProperty( $propertyId );
+			$this->propertyIds[$serializedId] = $dataTypeId;
+		}
+
+		return $this->propertyIds[$serializedId];
+	}
+
+}
--- a/tests/unit/Lookup/InProcessCachingDataTypeLookupTest.php
+++ b/tests/unit/Lookup/InProcessCachingDataTypeLookupTest.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Wikibase\Lib\Tests\Store;
+
+use PHPUnit_Framework_TestCase;
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Services\Lookup\InProcessCachingDataTypeLookup;
+
+/**
+ * @covers Wikibase\DataModel\Services\Lookup\InProcessCachingDataTypeLookup
+ *
+ * @licence GNU GPL v2+
+ * @author Katie Filbert < aude.wiki@gmail.com >
+ */
+class InProcessCachingDataTypeLookupTest extends PHPUnit_Framework_TestCase {
+
+	public function testIsMatchingDataTypeInProcessCaching() {
+		$propertyDataTypeLookup = $this->getMock(
+			'Wikibase\DataModel\Services\Lookup\PropertyDataTypeLookup'
+		);
+
+		$propertyDataTypeLookup->expects( $this->once() )
+			->method( 'getDataTypeIdForProperty' )
+			->with( new PropertyId( 'P1' ) )
+			->will( $this->returnValue( 'string' ) );
+
+		$cachingDataTypeLookup = new InProcessCachingDataTypeLookup( $propertyDataTypeLookup );
+
+		$dataType = $cachingDataTypeLookup->getDataTypeIdForProperty( new PropertyId( 'P1' ) );
+		$dataType = $cachingDataTypeLookup->getDataTypeIdForProperty( new PropertyId( 'P1' ) );
+
+		$this->assertEquals(
+			'string',
+			$dataType,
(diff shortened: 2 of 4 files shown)


tests/unit/Lookup/PropertyDataTypeMatcherTest.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @licence GNU GPL v2+
+ * @author Thiemo Mättig
+ */
+class PropertyDataTypeMatcherTest extends PHPUnit_Framework_TestCase {
+
+	/**
+	 * @dataProvider propertyDataTypeProvider
+	 */
+	public function testX( $propertyId, $dataType, $expected ) {
```
Benestar: testX? I'd say testIsMatchingDataType

tests/unit/Lookup/PropertyDataTypeMatcherTest.php
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	public function propertyDataTypeProvider() {
+		$p1 = new PropertyId( 'P1' );
+
+		return array(
+			array( $p1, 'equal', true ),
+			array( $p1, 'not equal', false ),
+		);
+	}
```
Benestar, line 59: Just some thoughts: In general, dataprovides are neat if there is a lot of data but for two test cases I think separate test methods are more appropriate.

Benestar:
Cool, seems to be helpful. Where exactly do you plan to use this in Wikibase?

tests/unit/Lookup/PropertyDataTypeMatcherTest.php
filbertkm: gah, have a bad version from the evil chain of patches :(

filbertkm:
@Benestar need it for all the places we examine snaks and statements in order to add stuff to parser output.

we keep checking if a property is, for example, commonsMedia and then add image links to parser output, then will need to check if it is coordinate then add geoData to parser output, etc.

https://gerrit.wikimedia.org/r/#/c/243613/

filbertkm:
test failure is unrelated in php 7:

https://github.com/wmde/WikibaseDataModelServices/issues/76

(still should be fixed, though)

JeroenDeDauw:
This does two things: adds caching to a PropertyDataTypeLookup and adapts a PropertyDataTypeLookup to an interface that does comparison. I'd expect the former to be done by a simple Decorator. It's not clear to me why the later is needed by looking at the linked commit on Gerrit. Can that part not simple be done by its two users? If not, then having a class like this one, though without the caching, seems sensible.

JeroenDeDauw:
Needs rebase - failure caused by something fixed already, which is why one one of the Travis builds is failing.

filbertkm:
@JeroenDeDauw it's more than two uses, but suppose a caching decorator would be nice

src/Lookup/InProcessCachingDataTypeLookup.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * PropertyDataTypeLookup that provides in-process caching.
+ *
+ * @since 3.1
+ *
+ * @licence GNU GPL v2+
+ * @author Katie Filbert < aude.wiki@gmail.com >
+ */
+class InProcessCachingDataTypeLookup implements PropertyDataTypeLookup {
```
thiemowmde, line 15: To what does "in process" refer to? I can not find an other class with "...InProcess..." in the name in my code base. Not in core, not in any extension. Isn't `CachingPropertyDataTypeLookup` better?

tests/unit/Lookup/InProcessCachingDataTypeLookupTest.php
```diff
@@ -0,0 +1,39 @@
...
+		);
+
+		$propertyDataTypeLookup->expects( $this->once() )
+			->method( 'getDataTypeIdForProperty' )
+			->with( new PropertyId( 'P1' ) )
+			->will( $this->returnValue( 'string' ) );
+
+		$cachingDataTypeLookup = new InProcessCachingDataTypeLookup( $propertyDataTypeLookup );
+
+		$dataType = $cachingDataTypeLookup->getDataTypeIdForProperty( new PropertyId( 'P1' ) );
```
thiemowmde, line 29: The duplicated lines do nothing. Either add a second assertion or remove one line.

src/Lookup/InProcessCachingDataTypeLookup.php
JeroenDeDauw, line 15: For an `CachingPropertyDataTypeLookup`, I'd expect it to not contain a specific caching strategy, and instead utilize a caching service.

~~We are using `InMemoryFoobarLookup` for this elsewhere. I think this name is slightly better than `InProcess`.~~

thiemowmde, line 15: Wow, so names **do** matter a lot. By typing "InMemory" I instantly found an existing [InMemoryDataTypeLookup](https://github.com/wmde/WikibaseDataModelServices/blob/master/src/Lookup/InMemoryDataTypeLookup.php) class ~~that already does what this proposal does~~.

JeroenDeDauw, line 15: ~~Oh yeah, that's actually the same interface. #derp~~

thiemowmde, line 15: Sorry, InMemoryDataTypeLookup is not about caching. What about a CachingPropertyDataTypeLookup that gets a HashBagOStuff via constructor? Isn't this exactly what the BagOStuff is about?

JeroenDeDauw, line 15: Shows I did not get enough sleep yesterday :)

Having a CachingPropertyDataTypeLookup that takes a caching interface (cannot be HashBagOStuff) would be nice, though I think having a specific decorator as done here is also fine.

filbertkm, line 15: I thought of using HashBagOStuff, but then we would have to put the CachingPropertyDataTypeLookup in WikibaseLib, since we then have a dependency on core.

tests/unit/Lookup/InProcessCachingDataTypeLookupTest.php
filbertkm, line 29: if PropertyDataTypeLookup::getDataTypeIdForProperty is called again in the second call, then the caching is not working properly.

how else would you like to test that?

thiemowmde, line 29: Oh, the important bit in the test is the "once". This is fine. However, having two assertions (by the way, please prefer `assertSame` whenever possible) after each call would possibly be less confusing.

JeroenDeDauw, line 29: Yeah, as is, the first `$dataType` assignment still has no use. Which means that if the return value for the not-cached cache is broken, the test won't catch it. Better to have distinct tests for distinct behaviours. I'll have a go at this in a follow-up.

tests/unit/Lookup/InProcessCachingDataTypeLookupTest.php
```diff
@@ -0,0 +1,39 @@
...
+		$dataType = $cachingDataTypeLookup->getDataTypeIdForProperty( new PropertyId( 'P1' ) );
+
+		$this->assertEquals(
+			'string',
+			$dataType,
+			'P1 has string data type and non-caching PropertyDataTypeLookup invoked only once.'
```
JeroenDeDauw, line 35: This message is not accurate - this assertion does not test the invocation of the decorated PropertyDataTypeLookup

PR #78, opened 2015-10-12
Update PHPCS rules

JeroenDeDauw:
- Allow placing operators on new lines
- Disallow using " for strings that do not have vars

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -72,11 +72,22 @@
     <rule ref="Squiz.Functions.FunctionDuplicateArgument" />
     <rule ref="Squiz.Functions.GlobalFunction" />
     <rule ref="Squiz.Scope" />
+
+
+    <rule ref="Squiz.Strings.DoubleQuoteUsage" />
+    <rule ref="Squiz.Strings.DoubleQuoteUsage.ContainsVar">
+        <severity>0</severity>
+    </rule>
+
     <rule ref="Squiz.WhiteSpace.CastSpacing" />
     <rule ref="Squiz.WhiteSpace.LanguageConstructSpacing" />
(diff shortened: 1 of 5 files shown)


phpcs.xml
```diff
@@ -72,11 +72,22 @@
     <rule ref="Squiz.Functions.FunctionDuplicateArgument" />
     <rule ref="Squiz.Functions.GlobalFunction" />
     <rule ref="Squiz.Scope" />
+
+
+    <rule ref="Squiz.Strings.DoubleQuoteUsage" />
+    <rule ref="Squiz.Strings.DoubleQuoteUsage.ContainsVar">
```
thiemowmde: Do you have a specific use case for this in mind? There is none in the current code as far as I can tell.

phpcs.xml
```diff
@@ -72,11 +72,22 @@
...
+    </rule>
+
     <rule ref="Squiz.WhiteSpace.CastSpacing" />
     <rule ref="Squiz.WhiteSpace.LanguageConstructSpacing" />
     <rule ref="Squiz.WhiteSpace.LogicalOperatorSpacing" />
 
-    <rule ref="Squiz.WhiteSpace.OperatorSpacing" />
+    <rule ref="Squiz.WhiteSpace.OperatorSpacing">
+        <properties>
+            <property name="ignoreNewlines" value="true" />
```
thiemowmde, line 88: Do you have a specific use case for this in mind? There is none in the current code as far as I can tell.

JeroenDeDauw, line 88: See https://github.com/squizlabs/PHP_CodeSniffer/issues/348

PR #80, opened 2015-10-13
Split test method in InProcessCachingDataTypeLookupTest

JeroenDeDauw:


tests/unit/Lookup/InProcessCachingDataTypeLookupTest.php
```diff
@@ -14,25 +14,38 @@
...
+
+	public function testWhenValueInCache_cacheValueIsReturned() {
+		$decoratedLookup = $this->getMock( 'Wikibase\DataModel\Services\Lookup\PropertyDataTypeLookup' );
+
+		$decoratedLookup->expects( $this->once() )
+			->method( 'getDataTypeIdForProperty' )
+			->will( $this->returnValue( 'string' ) );
+
+		$cachingLookup = new InProcessCachingDataTypeLookup( $decoratedLookup );
+		$cachingLookup->getDataTypeIdForProperty( new PropertyId( 'P1' ) );
```
thiemowmde, line 41: I still find it odd that this return value is not checked.

PR #81, opened 2015-10-13
Move PropertyDataTypeMatcher out of Lookup NS

JeroenDeDauw:


thiemowmde:
What's the motivation? I can see that "Lookup" is not a perfect match, but how is "Entity" better?

JeroenDeDauw:
The Entity NS has services that do something with an Entity and do not fit in a more specific NS. Which means this service fits there. It does not fit in Lookup, as it is not a lookup. Of course, if you find a more specific NS to put this into, then that is likely the better approach. Me and @filbertkm considered making a Property NS but decided against it.

thiemowmde:
Thanks for the insight.

PR #73, opened 2015-10-05
Add all missing PHPDoc tags

thiemowmde:
I find PHPDoc tags for class properties extremely helpful, if not critical, because it saves one from scanning the whole file for the intended type.

I'm _not_ adding tags in cases where it's really, really obvious from the method signature.

PR #82, opened 2015-10-14
Fix type hint in ByPropertyIdGrouper

adrianheine:
indexPropertyIdProviders is called from __construct with
`PropertyIdProvider[]|Traversable`.

### DIFF
--- a/phpmd.xml
+++ b/phpmd.xml
@@ -8,9 +8,6 @@
         <exclude name="ExcessiveClassComplexity" />
         <exclude name="TooManyPublicMethods" />
     </rule>
-    <rule name="rulesets/codesize.xml/TooManyPublicMethods">
-        <exclude-pattern>tests/unit/</exclude-pattern>
-    </rule>
 
     <rule ref="rulesets/controversial.xml">
         <exclude name="CamelCaseMethodName" />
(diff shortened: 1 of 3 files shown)


thiemowmde:
So I did that by mistake in #73, but no test covered it? Whoops. I added tests. Feel free to merge, +2 from my side.

thiemowmde:
PHPMD complains: `Cannot load source file for class:` ... That's all it says. What class?

I give up. `<exclude-pattern>` does not work as I expect.

JeroenDeDauw:
@thiemowmde Thanks a lot for adding the tests. However, if you put commits on top of someones PR, please don't leave it in such a messy state. This makes moving the thing forward rather awkward for me. Now we have a failing commit in the history, for which I'd need to amend your commits on yet someones elses PR...

thiemowmde:
> don't leave it in such a messy state.

What's "messy" in 112b3b7e2ee0360c774a3ed0fbc47d0e41a098a2? I left 6d6e21ebc6b7c9a4832abde682e87af1923a1e64 so you can see what I unsuccessfully tried. Feel free to squash it away if it disturbs something (but I really do not understand why and where such a commit can be a problem, please tell me).

> I'd need to amend your commits

Not sure what you mean. You can squash my two commits together and then either re-enable the rule in an other pull request or by adding another commit on top of Adrian's and mine, as you did.

PR #90, opened 2015-11-19
Introduce FormatterUrlLookup.

brightbyte:
NOTE: needs a minor release. 

[Bug: T112762](http://phabricator.wikimedia.org/T112762)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,7 +1,7 @@
 # Wikibase DataModel Services release notes
 
-## Version 3.2 (dev)
-* Introduced PropertyInfoFormatter
+## Version 3.2.0 (dev)
+* Added `PropertyFormatterUrlLookup`
 
 ## Version 3.1.1 (2015-11-15)
 * Made `EntityRetrievingTermLookup` handle `EntityLookupException` ([T118581](https://phabricator.wikimedia.org/T118581))
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,8 @@
 # Wikibase DataModel Services release notes
 
+## Version 3.2 (dev)
+* Introduced PropertyInfoFormatter
+
 ## Version 3.1.1 (2015-11-15)
 * Made `EntityRetrievingTermLookup` handle `EntityLookupException` ([T118581](https://phabricator.wikimedia.org/T118581))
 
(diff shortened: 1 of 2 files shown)


thiemowmde:
We already know we need to lookups for very similar things, one for the URL of an identifier and one for the URI. Should these two be in the same interface? Two methods? One method with a parameter? Isn't a generic "check if property P1 does have a statement with property P2 and return it's value" lookup better?

brightbyte:
I'm inclined to add the method for getting the URI pattern to the same interface.
A generic interface that takes a PropertyID means that the caller has to somehow know the right property ID (we'd have to inject config variables). It's also unclear what the lookup should do if the requested property isn't configured to be recorded in the property_info table.

JeroenDeDauw:
Where do you intend to use this interface? And why put it in this component? (-1 until answered, not saying there is an issue before though.)

thiemowmde:
I'm also curious why this interface needs to be in this component. It seems to be very specific for a use case that no other user of this component may have. Can't this live in WikibaseRepo?

brightbyte:
The interface is here because it is needed by both Client and Repo. I could also put it into Lib, if you prefer. It's used here: https://gerrit.wikimedia.org/r/#/c/254161/

The interface is specific for providing URL patterns associated with properties. Any client that wants to generate URLs for external IDs will need this (unless we generate Snaks with derived values attached directly from the JSON the client gets from the API - but even that would only work for clients that use the web API, not for client wikis).

JeroenDeDauw:
> It's used here: https://gerrit.wikimedia.org/r/#/c/254161/

That does not answer my question actually. Implementing an interface is not really making use of it. Where will we have something of the type defined by this interface?

JeroenDeDauw:
> The interface is here because it is needed by both Client and Repo.

Right. Thanks for considering that. With the info I have so far I think it's a better tradeoff to keep it in Lib with the classes implementing it. That is more cohesive and keeps the "formatter URL pattern" concept out of this component.

thiemowmde:
I agree with Jeroen, see my comments in https://gerrit.wikimedia.org/r/254161.

brightbyte:
Ok, I'll move this into Lib. I was following the example of PropertyDataTypeLookup. Do you think that would also better be in Lib?

In any case, we should move the formatters  into their own component. And the service implementation they use also need a new component, since most of them depend on MediaWiki.

JeroenDeDauw:
> I was following the example of PropertyDataTypeLookup. Do you think that would also better be in Lib?

No - it has implementations in this component, is used outside of Wikibase.git and does not introduce new concepts to this component.

PR #91, opened 2015-11-19
Introduce StatementGrouper and StatementFilter

thiemowmde:
We currently have and will have quite some use cases where we want to group statements based on different criteria:
- Grouping statements with the same property id is something we already do. Currently this is done in `StatementGroupListView::groupStatementsByProperties`. We also have ByPropertyIdArray and ByPropertyIdGrouper (by @Benestar), which are both almost unused and could be replaced with the ByPropertyIdStatementGrouper in this patch.
- We want to group statements based on their properties data type, namely an "Identifiers" section.
- We may want to group statements based on a configured set of property ids. An example for that would be a "Biography" group with properties like "date of birth", "place of birth" and so on.
- We may want to have groups per rank.

The StatementFilter approach suggested in [T117421](https://phabricator.wikimedia.org/T117421#1807403) allows to combine these criteria, e.g. have an "Identifiers" group via data type, a "Biography" group via a set of property ids and a "Statements" group for the rest.

[Bug: T118950](https://phabricator.wikimedia.org/T118950)
[Bug: T118951](https://phabricator.wikimedia.org/T118951)


JeroenDeDauw:
If this is to solve some problem in the view code, I recommend writing the code there in a manner that solves the problem, and then looking at what you can move here. When what you need exactly is still fuzzy, adding it in another component is risky and painful (including for the reviewer if insufficient context is provided).

> Tests are missing

I also suggest writing the tests first. If you write them now, I'm not going to have a lot of confidence in them.

src/Statement/Grouper/StatementGrouper.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @licence GNU GPL v2+
+ * @author Thiemo Mättig
+ */
+interface StatementGrouper {
+
+	/**
+	 * @param StatementList $statements
+	 *
+	 * @return StatementList[] An associative array, mapping statement group identifiers to
```
JeroenDeDauw: Some further explanation of what "statement group identifiers" are would be nice

thiemowmde: Whatever PHP allows as array keys. Usually a string key, but can also be an integer. I'm not sure what I should write to make this more obvious.

tests/unit/Statement/Grouper/NullStatementGrouperTest.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @covers Wikibase\DataModel\Services\Statement\Grouper\NullStatementGrouper
+ *
+ * @licence GNU GPL v2+
+ * @author Thiemo Mättig
+ */
+class NullStatementGrouperTest extends PHPUnit_Framework_TestCase {
+
+	public function test() {
```
JeroenDeDauw: I thought PHPUnit would get mad if you just called a method test. Perhaps not. Not sure now it will actually run it. Better to give the test a name :)

thiemowmde: PHPUnit runs this. I usually start with `test()` and name it when it becomes obvious what it tests.

This is a mistake. I fixed it and a similar one in the other grouper.

src/Statement/Grouper/StatementGrouper.php
JeroenDeDauw: Not the types, but where the values come from. When I first read this and had not looked at the code, I did not know what this was at all

JeroenDeDauw:
+1

JeroenDeDauw:
+1 (+2 if someone who is up to speed with the code where this is intended to be used +1's)

src/Statement/Grouper/StatementGrouper.php
thiemowmde: > where the values come from.

This is an interface. It can not know where this comes from. May be from some constructor parameter, may be some generated keys, depending on the implementation.

JeroenDeDauw: Is there something these identifiers have in common? It's still not clear to me what as a user of this interface I can expect. It's possible they do not have anything in common, and that if I get passed an arbitrary implementation of the interface, I can make no assumptions about these keys. If that is the case, it is not bad, though then naming the concept might be confusing, as it somewhat implies it has relevant attributes to the user of the interface.

> This is an interface.

obvious

thiemowmde: I really try, but all your comments do is confusing me. I neither understand what I should do nor what you are missing. You can see the code. It's all here. Why do your questions sound like you haven't looked at it? What's the goal of these questions? Teaching me something? But what? Please make suggestions or ask specific questions I can answer without getting sarcastic comments as a response.

JeroenDeDauw: The goal of these questions is for me to understand something which I currently do not. "Is there something these identifiers have in common" is a specific question.

Interfaces are much closer related to their users than to their implementations. The question I'm asking is about intend and cannot be answered just by looking at the implementations.

Besides, one should not have to look at implementations to be able to understand an interface.

JonasKress: @JeroenDeDauw  It is not voodoo it is just a key to identify the group.  More background inside of the ticket: https://phabricator.wikimedia.org/T117421

src/Statement/Grouper/FilteringStatementGrouper.php
```diff
@@ -0,0 +1,84 @@
...
+	private $filters = array();
+
+	/**
+	 * @var string
+	 */
+	private $defaultGroupIdentifier = 'statements';
+
+	/**
+	 * @param array $filters An associative array, mapping statement group identifiers to either
+	 *  filters, or to null for the default group.
```
JonasKress: filters should link to the StatementFilter interface, so it is clear how to provide a filter

src/Statement/Grouper/FilteringStatementGrouper.php
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	private function setFilters( array $filters ) {
+		foreach ( $filters as $key => $filter ) {
+			$this->groups[$key] = new StatementList();
+
+			if ( $filter === null ) {
+				$this->defaultGroupIdentifier = $key;
+			} else {
+				$this->filters[$key] = $filter;
```
JonasKress: Check for StatementFilter interface

src/Statement/Grouper/FilteringStatementGrouper.php
```diff
@@ -0,0 +1,84 @@
...
+
+		return $this->groups;
+	}
+
+	/**
+	 * @param Statement $statement
+	 *
+	 * @return string Statement group identifier
+	 */
+	private function getKey( Statement $statement ) {
```
JonasKress, line 103: Rename to getGroupKey?

src/Statement/Grouper/StatementGrouper.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Wikibase\DataModel\Services\Statement\Grouper;
+
+use Wikibase\DataModel\Statement\StatementList;
+
+/**
+ * @since 3.2
+ *
+ * @licence GNU GPL v2+
+ * @author Thiemo Mättig
```
JonasKress, line 11: Please describe what the interface is for.

src/Statement/Filter/StatementFilter.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Wikibase\DataModel\Services\Statement\Filter;
+
+use Wikibase\DataModel\Statement\Statement;
+
+/**
+ * @since 3.2
+ *
+ * @licence GNU GPL v2+
+ * @author Thiemo Mättig
```
JonasKress: Please add a description for this interface.

JonasKress:
Good job @thiemowmde !

src/Statement/Grouper/StatementGrouper.php
thiemowmde, line 11: It groups statements, just as the name of the interface says. I will not add stuff like "use the StatementGrouper interface to group statements" as a description.

src/Statement/Filter/StatementFilter.php
thiemowmde: It filters statements. There is only 1 public method. It requires 1 parameter and returns a bool. I do not know how more obvious it can be. What would a description add to this? What do you suggest?

src/Statement/Grouper/FilteringStatementGrouper.php
thiemowmde, line 103: getGroupIdentifier would be more accurate, but way to long. Short scope, short name, remember?

src/Statement/Grouper/FilteringStatementGrouper.php
thiemowmde: "Link" how? This can not be `StatementFilter[]`, it must be `array`.

JonasKress: @see StatementFilter ?

src/Statement/Grouper/StatementGrouper.php
thiemowmde: > "Is there something these identifiers have in common" is a specific question.

To what does "common" refer to? What do you expect? A common prefix or something? A predefined set of identifiers? They are nothing but array keys, as I already explained. Either strings or integers. And since this is an interface it can **not** know where this comes from. How could it? Why should it? What else could I say?

We two are going in circles, again and again, and it starts driving me crazy. You asking the same obscure questions again and again and me explaining the same obvious facts again and again. I would love to understand why this happens. I would love to avoid this, preferably not by simply ignoring what you write.

> The question I'm asking is about intend

What do you mean with "intend"? This is an interface to group statements, a.k.a. splitting a StatementList into multiple StatementLists. That's all. How this is done depends on the implementation and can not be described in the interface. How would such a description look like?

There. I did it again. I repeated myself. I spend an other hour on nothing but trying to answer a comment I still do not understand.

thiemowmde:
I just realized there is a [StatementFilter in WikibaseDataModel](https://github.com/wmde/WikibaseDataModel/blob/master/src/Statement/StatementFilter.php), introduced in https://github.com/wmde/WikibaseDataModel/pull/525 and still completely unused in production code. Not sure what to do about this. I do have the feeling this should be a service and not in the core data model.

src/Statement/Grouper/StatementGrouper.php
JeroenDeDauw: @Jonas-WMDE so for the user of this interface this key has no special meaning beyond it being an array key?

src/Statement/Grouper/FilteringStatementGrouper.php
JonasKress, line 103: It's the opposite short scope long name ;)

src/Statement/Grouper/StatementGrouper.php
JonasKress: In that scope yes. In view the key will be used as part of the message key.

src/Statement/Grouper/FilteringStatementGrouper.php
```diff
@@ -0,0 +1,101 @@
...
+	 *  StatementFilter objects, or to null for the default group.
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function __construct( array $filters ) {
+		$this->setFilters( $filters );
+	}
+
+	/**
+	 * @param array $filters
```
adrianheine, line 37: `(StatementFilter|null)[]`

adrianheine:
Discussion with @Jonas-WMDE @JeroenDeDauw lead to the decision to use `Wikibase\DataModel\Statement\StatementFilter` and remove `Wikibase\DataModel\Services\Statement\Filter\StatementFilter` from this pull request.

src/Statement/Grouper/FilteringStatementGrouper.php
JeroenDeDauw, line 37: Unfortunately PHP tools do not understand that. What is currently done is the best you can do.

And yeah, that's really sad. As is the rest of PHPs totally broken collection support :(

src/Statement/Filter/PropertySetStatementFilter.php
```diff
@@ -0,0 +1,40 @@
...
+
+	/**
+	 * @var string[]
+	 */
+	private $propertyIds;
+
+	/**
+	 * @param string[]|string $propertyIds One or more property id serializations.
+	 */
+	public function __construct( $propertyIds ) {
```
adrianheine, line 24: Shouldn't this take an array of `PropertyId`s instead? We could create a fresh array with the serialization as key to reduce complexity of matching.

src/Statement/Grouper/ByPropertyIdStatementGrouper.php
```diff
@@ -0,0 +1,38 @@
...
+				$groups[$id] = new StatementList();
+			}
+
+			$groups[$id]->addStatement( $statement );
+		}
+
+		return $groups;
+	}
+
+}
```
adrianheine, line 38: I don't think there is a plan to use this class currently, is there?

src/Statement/Grouper/FilteringStatementGrouper.php
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @licence GNU GPL v2+
+ * @author Thiemo Mättig
+ */
+class FilteringStatementGrouper implements StatementGrouper {
+
+	/**
+	 * @var StatementList[]
+	 */
+	private $groups = array();
```
adrianheine: Why does this service have state? Doesn't seem necessary to me.

src/Statement/Filter/PropertySetStatementFilter.php
JeroenDeDauw, line 24: Might be a good use case for http://php.net/manual/en/class.splobjectstorage.php ?

src/Statement/Filter/PropertySetStatementFilter.php
```diff
@@ -0,0 +1,40 @@
...
+	 * @param Statement $statement
+	 *
+	 * @return bool
+	 */
+	public function statementMatches( Statement $statement ) {
+		$id = $statement->getPropertyId()->getSerialization();
+		return in_array( $id, $this->propertyIds );
+	}
+
+}
```
adrianheine, line 40: Probably not going to be used, too.

src/Statement/Filter/DataTypeStatementFilter.php
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	private $dataTypes;
+
+	/**
+	 * @param PropertyDataTypeLookup $dataTypeLookup
+	 * @param string[]|string $dataTypes One or more data type identifiers.
+	 */
+	public function __construct( PropertyDataTypeLookup $dataTypeLookup, $dataTypes ) {
+		$this->dataTypeLookup = $dataTypeLookup;
+		$this->dataTypes = (array)$dataTypes;
```
brightbyte, line 34: flip it - so we can do "set style" lookups for performance.

src/Statement/Filter/PropertySetStatementFilter.php
brightbyte, line 24: Yes, this should take PropertyIds

src/Statement/Grouper/FilteringStatementGrouper.php
```diff
@@ -0,0 +1,113 @@
...
+use Wikibase\DataModel\Statement\StatementFilter;
+use Wikibase\DataModel\Statement\StatementList;
+
+/**
+ * @since 3.2
+ *
+ * @licence GNU GPL v2+
+ * @author Thiemo Mättig
+ */
+class FilteringStatementGrouper implements StatementGrouper {
```
brightbyte, line 16: Why do we need statement groupers at all? Wouldn't it be sufficient to give the view code a filter?
E.g. instead of creating two groups for "ID statements" and "non-ID statements", just call the view code once with a filter that matches all ID statements, and once with a filter that matches all non-ID statements?

Also, I propose to change our internal code to work on per-property statement groups everywhere. We shouw statements grouped by property, and we group statements be property in the JSON representation. Wwe should probably also do this consistently in PHP. It's just needlessly painful to work on the flat statement list.

src/Statement/Filter/DataTypeStatementFilter.php
adrianheine, line 34: ack, that would be better. Or, as Jeroen suggested below, use http://php.net/manual/en/class.splobjectstorage.php.

src/Statement/Grouper/FilteringStatementGrouper.php
adrianheine, line 16: Funny, I'm currently considering removing the grouping from the JS code and the JSON in `wbEntity`. To me, it's a pain to work with the grouped statement list, since we are now going to look at them by GUID alone.

As for statement groupers: The fact that this class is non-trivial seems convincing enough to me. Note that using non-exclusive filters for grouping is not that obvious or trivial. We might just as well have groupers that don't work with filters.

src/Statement/Filter/PropertySetStatementFilter.php
maettig: Whut? Why? How would this look like in production?
```php
$setting = array( 'P1', 'P2' );
$objects = $this->pointlessConversionToObects( $setting );
// Pass to the filter
$strings = $this->pointlessConversionBackToStrings( $objects );
// Use for filtering
```
Can we please **not** do this?

src/Statement/Grouper/ByPropertyIdStatementGrouper.php
maettig: Sure it is. We already do this, hardcoded in a private method.

src/Statement/Grouper/FilteringStatementGrouper.php
maettig: I'm sorry? I assure you this is not a mistake. I spend hours and hours on this.

src/Statement/Filter/PropertySetStatementFilter.php
maettig: Erm, no? Why do you say that? Did you read the commit message? Did you read the related tasks descriptions?

src/Statement/Filter/DataTypeStatementFilter.php
maettig: No. Heck, what are you guys doing to my code? I **benchmarked** this! An array_flip is completely pointless in this class. It's completely pointless in almost all cases. Rule of thumb: Do it when the array usually contains more than 100 strings. This is not the case here.

src/Statement/Grouper/FilteringStatementGrouper.php
maettig: See my comment in https://gerrit.wikimedia.org/r/#/c/254394/3/view/src/ItemView.php.

src/Statement/Filter/PropertySetStatementFilter.php
adrianheine, line 24: Where would the conversion to string happen?

thiemowmde:
I regret I have looked at this. I should recover instead. Bugs introduced, unnecessary changes, and comments that are, at best, written from an uninformed perspective, and at worst unrelated. I wrote this code 12 days ago, in pair with @Jonas-WMDE. I got reviews from @JeroenDeDauw and @Jonas-WMDE. And now that I'm gone people are tearing my code apart? Do you know how this feels?

src/Statement/Grouper/FilteringStatementGrouper.php
thiemowmde, line 16: See my comment in https://gerrit.wikimedia.org/r/#/c/254394/3/view/src/ItemView.php.

src/Statement/Filter/DataTypeStatementFilter.php
thiemowmde, line 34: No. Heck, what are you guys doing to my code? I **benchmarked** this! An array_flip is completely pointless in this class. It's completely pointless in almost all cases. Rule of thumb: Do it when the array usually contains more than 100 strings. This is not the case here.

src/Statement/Filter/PropertySetStatementFilter.php
thiemowmde, line 40: Erm, no? Why do you say that? Did you read the commit message? Did you read the related tasks descriptions?

src/Statement/Grouper/ByPropertyIdStatementGrouper.php
thiemowmde, line 38: Sure it is. We already do this, hardcoded in a private method.

src/Statement/Filter/PropertySetStatementFilter.php
thiemowmde, line 24: Whut? Why? How would this look like in production?

``` php
$setting = array( 'P1', 'P2' );
$objects = $this->pointlessConversionToObects( $setting );
// Pass to the filter
$strings = $this->pointlessConversionBackToStrings( $objects );
// Use for filtering
```

Can we please **not** do this?

src/Statement/Grouper/ByPropertyIdStatementGrouper.php
adrianheine, line 38: What, where?

src/Statement/Filter/PropertySetStatementFilter.php
adrianheine, line 40: I wrote the tasks. We could use this, for sure, but there are no plans to actually do that.

src/Statement/Grouper/FilteringStatementGrouper.php
thiemowmde: I'm not happy with this too, because it means statements will stack up when calling this filter multiple times. Since I'm quite proud of the tests for this class, feel free to refactor it if you can do this without touching the tests.

adrianheine:
@thiemowmde I'm sorry how you feel about  that. Let's talk about it when you get back, shall we?

src/Statement/Grouper/FilteringStatementGrouper.php
adrianheine: I think your tests will still pass with what @Jonas-WMDE did.

src/Statement/Grouper/ByPropertyIdStatementGrouper.php
thiemowmde, line 38: I'm sure I mentioned this in the commit message and indeed, I did.

src/Statement/Grouper/FilteringStatementGrouper.php
thiemowmde: What did Jonas do? I'm confused. The additional commit does have your avatar, not Jonas'.

adrianheine: Jonas amended my change.

src/Statement/Filter/PropertySetStatementFilter.php
thiemowmde, line 40: Without the most basic implementation(s) it's almost pointless to put these interfaces here, in a component other users can and should use. If you plan to remove everything that's not going to be used in Wikibase right away, with this being the only reason, then please abandon this patch completely and implement what's needed in WikibaseView. You can most probably ditch all the interfaces then and use the implementations directly. Or even go with @brightbyte's https://github.com/wmde/WikibaseDataModelServices/pull/91#discussion_r46144648.

src/Statement/Grouper/FilteringStatementGrouper.php
thiemowmde, line 16: There **are** groupers that don't work with filters. This is one of the main reasons why I added the most basic groupers to this patch, even if they are not relevant for the current story. Also see https://github.com/wmde/WikibaseDataModelServices/pull/91/files#r46273418.

src/Statement/Grouper/FilteringStatementGrouper.php
thiemowmde: GitHub shows a tiny avatar on top of the original avatar when somebody amends a commit done by somebody else. This is not the case here. How is this possible? What did Jonas do?

JeroenDeDauw:
@brightbyte did the concerns all get addressed, or is there need for follow ups?

thiemowmde:
Why is this now merged with all the issues unsolved? Do you plan to upload an other patch that fixes these issues? Are there tickets that track these issues? Or was this a mistake? I really, really do not understand.

Benestar:
I'm also a bit confused? Do we now need StatementGroupers or don't we need them?? There are still some threads open I think.

thiemowmde:
All remaining issues are solved in #94.

PR #92, opened 2015-11-23
Add EntityDifferStrategy::canDiffEntityTypes

thiemowmde:
This is split from #61.

The fact that we currently do not support diffs between different types of entities should not be hardcoded so deep in the system. I believe that what I do in this patch makes the code easier to understand. Look at this line:

``` php
$this->getDiffStrategy( $from->getType() )->diffEntities( $from, $to )
```

It looks suspicious. How can the strategy be based on **one** side, when the diff does have two? This patch fixes this.

### DIFF
--- a/src/Diff/EntityDiffer.php
+++ b/src/Diff/EntityDiffer.php
@@ -39,7 +39,8 @@ class EntityDiffer {
 	public function diffEntities( EntityDocument $from, EntityDocument $to ) {
 		$this->assertTypesMatch( $from, $to );
 
-		return $this->getDiffStrategy( $from->getType() )->diffEntities( $from, $to );
+		$differ = $this->getDiffStrategy( $from->getType(), $to->getType() );
+		return $differ->diffEntities( $from, $to );
 	}
 
 	private function assertTypesMatch( EntityDocument $from, EntityDocument $to ) {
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
I think it's entirely fine for these strategies to only be able to diff entities of the same type, if that's all that is needed. That said, there is nothing wrong with creating a differ that can diff entities of different types if we need that somewhere. Did we pick up a feature that needs this? Or is this subjectively "fixing" something for the sake of it?

I remember us talking about the previous PR and coming to the conclusion there is no need to implement the straggly interface in the differ it adds. If that is the case, then why is the change to the interface needed?

PS: I'm a bit annoyed that I need to repeat myself so often due to the creation of new PRs. Would you be fine with using a single one in such cases in the future?

RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel Services release notes
 
+## Version 3.2.0 (dev)
+* Removed `EntityDifferStrategy::canDiffEntityType`
```
JeroenDeDauw: This cannot go into 3.2 as it is a breaking change

thiemowmde:
This is a different pull request. It fixes a different issue.

thiemowmde:
As I already explained in https://github.com/wmde/WikibaseDataModelServices/pull/61#issuecomment-137174665 this is not about adding a new feature "we need somewhere". This is about removing a barrier from the code that just does not need to be there. Everything works just fine without this code actively blocking certain attempts to use diff.

That said, I could also ask when we pick up a feature that requires blacklisting certain usages of diff?

JeroenDeDauw:

PR #98, opened 2015-12-09
Rename $autoLoader to $classLoader

thiemowmde:
I find it helpful when the var name gives a hint that it is of type `ClassLoader`.

Also see:
- https://github.com/wmde/WikibaseDataModel/pull/582
- https://github.com/wmde/WikibaseInternalSerialization/pull/89


JeroenDeDauw:
This stuff is called autoloader in PHP world and we've been using this var name everywhere since forever. Not going to argue the point though as I think we have better use for our time.

thiemowmde:
I'm touching this because it is **not** what we are doing "everywhere".

addshore:
No strong opinion either way, but we do seem to have it called $classLoader in other places.

PR #99, opened 2015-12-09
Avoid full qualified class names

thiemowmde:


JeroenDeDauw:
I think the old stuff is just fine. Not going to argue the point as I think we have better thing to do with our time.

thiemowmde:
So why are you arguing then and not just merging this?

PR #100, opened 2015-12-15
Escaping formatter doesn't always return HTML

thiemowmde:
I think this qualifies as a bug in the documentation.


JeroenDeDauw:
Weird class (name), nothing in it is specific to escaping.

JeroenDeDauw:
+1

PR #101, opened 2015-12-15
Re-enable two disabled PHPMD rules

thiemowmde:
This patch re-enables two PHPMD rules that are currently disabled entirely: ExcessiveClassComplexity, and ExcessiveMethodLength. The default "minimum" for ExcessiveMethodLength is 100. ItemDiffTest breaks this, that's why I set it to 139 for now.


JeroenDeDauw:
Re-running tests

PR #102, opened 2015-12-15
PSR-4 compatible namespaces for test classes

thiemowmde:


tests/unit/DataValue/ValuesFinderTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace Wikibase\Lib\Test;
```
JeroenDeDauw, line 3: :D

PR #105, opened 2015-12-28
Fix inconsistent StatementGuid creation and parsing

thiemowmde:
There never was a test for a statement id with more than one dollar sign. Neither a positive nor a negative test. Therefore I consider this a bug. It's allowed to create a statement id "Q1$$" (where the first dollar sign is the separator), so it should be possible to parse this.

### DIFF
--- a/src/Statement/StatementGuidParser.php
+++ b/src/Statement/StatementGuidParser.php
@@ -7,6 +7,11 @@ use Wikibase\DataModel\Entity\EntityIdParsingException;
 use Wikibase\DataModel\Statement\StatementGuid;
 
 /**
+ * A parser capable of splitting a statement id into the entity id of the entity the statement
+ * belongs to, and the randomly generated global unique identifier (GUID).
+ *
+ * @see StatementGuid
+ *
  * @since 1.0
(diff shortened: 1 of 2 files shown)


src/Statement/StatementGuidParser.php
```diff
@@ -7,6 +7,11 @@
 use Wikibase\DataModel\Statement\StatementGuid;
 
 /**
+ * A parser capable of splitting a statement id into the entity id of the entity the statement
+ * belongs to, and the randomly generated global unique identifier (GUID).
+ *
+ * @see Wikibase\DataModel\Statement\StatementGuid
```
Benestar: No need for the fully qualified name as this is already imported some lines above.

Benestar:
+1 though I think we maybe should forbid an empty GUID in `StatementGuid` itself.

thiemowmde:
This patch is not about empty strings. Not at all.

I can see where the idea comes from. I also added a lot of `=== ''` checks to many classes over the past year. But in this case I disagree. You see, why should `Q1$` be invalid? It's not empty.

PR #111, opened 2016-02-16
Unify duplicate getSiteLinksInDiffFormat methods

thiemowmde:
This patch:
- Narrows an interface from Item to SiteLinkList.
- Makes use of toArray.
- Renames methods and variables to make the two implementations much more identical. This is duplicate code, but I do not think we must get rid of this. But I find it helpful when it's a bit more obvious what's going on (think of a fulltext search for "getSiteLinksInDiffFormat").


JeroenDeDauw:
I was confused by the PR message, as no unification happens

thiemowmde:
Harmonize. I'm always confusing the words.

PR #114, opened 2016-02-19
Add RestrictedEntityLookup::resetEntityAccessCount

mariushoch:
Needed to reset it between parser runs and things like that.

### DIFF
--- a/src/Lookup/RestrictedEntityLookup.php
+++ b/src/Lookup/RestrictedEntityLookup.php
@@ -105,6 +105,15 @@ class RestrictedEntityLookup implements EntityLookup {
 	}
 
 	/**
+	 * Resets the number of entities loaded via this object.
+	 *
+	 * @since 3.2.1
+	 */
+	public function resetEntityAccessCount() {
+		$this->entityAccessCount = 0;
+	}
+
+	/**
 	 * Whether an entity has been accessed before via this RestrictedEntityLookup.
 	 *
 	 * @since 2.0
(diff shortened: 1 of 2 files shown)


src/Lookup/RestrictedEntityLookup.php
```diff
@@ -105,6 +105,15 @@ public function getEntityAccessCount() {
 	}
 
 	/**
+	 * Resets the number of entities loaded via this object.
+	 *
+	 * @since 3.2.1
```
JeroenDeDauw, line 110: This is wrong. Fixed by #115

mariushoch, line 110: It's not, I plan to release a 3.2.1 based on this for backporting.

JeroenDeDauw, line 110: This is not a bugfix, you cannot backport it to a bugfix release

mariushoch, line 110: I don't know how we should handle these situations. I would like to avoid having to locally apply patches to the build for minimal updates… and re-tagging a minimal 3.3.0 release also sounds awry to me.

JeroenDeDauw, line 110: I'm not sure what you mean with "having to locally apply patches to the build for minimal updates".

> and re-tagging a minimal 3.3.0 release also sounds awry to me.

semver does not allow retagging in any siatuation, not even when you accidentally have a breaking change in a minor release. I'm not sure why you'd need to retag...

thiemowmde:
Reverted in #117. Discussion continues there.

PR #117, opened 2016-02-21
Introduce RestrictedEntityLookup::reset

mariushoch:
And remove the reset access count method, as that didn't reset the list of entities loaded, thus serves no real purpose.

[Bug: T127462](http://phabricator.wikimedia.org/T127462)

### DIFF
--- a/src/Lookup/RestrictedEntityLookup.php
+++ b/src/Lookup/RestrictedEntityLookup.php
@@ -105,12 +105,13 @@ class RestrictedEntityLookup implements EntityLookup {
 	}
 
 	/**
-	 * Resets the number of entities loaded via this object.
+	 * Resets the number and list of entities loaded via this object.
 	 *
 	 * @since 3.4
 	 */
-	public function resetEntityAccessCount() {
+	public function reset() {
 		$this->entityAccessCount = 0;
+		$this->entitiesAccessed = array();
 	}
 
 	/**
--- a/tests/unit/Lookup/RestrictedEntityLookupTest.php
+++ b/tests/unit/Lookup/RestrictedEntityLookupTest.php
@@ -67,14 +67,18 @@ class RestrictedEntityLookupTest extends \PHPUnit_Framework_TestCase {
 		$this->assertSame( 5, $lookup->getEntityAccessCount() );
 	}
 
-	public function testResetEntityAccessCount() {
+	public function testReset() {
 		$lookup = new RestrictedEntityLookup( $this->getEntityLookup(), 200 );
 		$lookup->getEntity( new ItemId( 'Q1' ) );
 
-		$lookup->resetEntityAccessCount();
+		$lookup->reset();
 
 		$lookup->getEntity( new ItemId( 'Q2' ) );
 		$this->assertSame( 1, $lookup->getEntityAccessCount() );
+
+		// An entity accessed before, but after reset counts again
+		$lookup->getEntity( new ItemId( 'Q1' ) );
+		$this->assertSame( 2, $lookup->getEntityAccessCount() );
 	}
 
 	public function testGetEntity() {


Benestar:
This is now a breaking change...

mariushoch:
> This is now a breaking change...

Technically, yes.

JeroenDeDauw:
How is this a breaking change? The change method was not present in any previous release.

thiemowmde:
This should have never been released, but I do see a 3.2.1 with this method released on Saturday, while the earlier (!) released 3.3.0 misses this method. What a mess. Personally I would delete the 3.2.1 tag, but since I'm not sure what's going on here, why this is so critical and why this bogus 3.2.1 tag was done with no review I'm aware of, I feel more like -2'ing everything going on in this component till we discussed this in person.

JeroenDeDauw:
> Personally I would delete the 3.2.1 tag

Please do not. If something was messed up in 3.2.1, then the semver way of fixing this is by releasing a 3.2.2. No deletion of tags, or modification of tags.

Benestar:
Ok, so lets do that. Do a correction release as 3.2.2 and then check if the functionality introduced in this pull request really serves the purpose of the ticket. If that is verified, we can do a 3.4.0 release and use that new functionality.

thiemowmde:
@JeroenDeDauw But the fact that 3.2.1 was tagged later (!) than 3.3.0 introduced (!) a breaking change into 3.3.0 that was not there before. This should have never happened. I consider this a mistake that must be undone. Releasing a 3.2.2 with nothing but a revert of that change is identical to simply deleting 3.2.1.

Benestar:
> What do I do if I accidentally release a backwards incompatible change as a minor version?
> 
> As soon as you realize that you’ve broken the Semantic Versioning spec, fix the problem and release a new minor version that corrects the problem and restores backwards compatibility. Even under this circumstance, it is unacceptable to modify versioned releases. If it’s appropriate, document the offending version and inform your users of the problem so that they are aware of the offending version.

We should definitely release a 3.2.2 per semver

thiemowmde:
Which means checking out 3.2.0 and tagging it as 3.2.2. @mariushoch, will you do this?

mariushoch:
> Which means checking out 3.2.0 and tagging it as 3.2.2. @mariushoch, will you do this?

Yes, I can do that if it's ok for you.

Benestar:
Great, thanks @mariushoch. Did you verify that the `reset` method introduced in this pull request does its job in fixing the linked issue?

mariushoch:
> Great, thanks @mariushoch. Did you verify that the `reset` method introduced in this pull request does its job in fixing the linked issue?

Yes, please see https://gerrit.wikimedia.org/r/272333 for that.

Benestar:
As `resetEntityAccessCount` isn't included in the previous 3.3 release this is no breaking change and can be released in a 3.4

thiemowmde:
The code in this patch is fine, but I will not merge it until it's clear how we solve all this.
- We can backport a Wikibase.git patch that uses the current 3.2.1 tag. From my point of view this would be the easiest solution. It should be set up so it does not use the 3.2.2 we need to release here. Is there something I'm missing?
- You can copy the file to Wikibase.git and use that. Done.
- We can merge this patch, release 3.4 and use that in the backport release you are planning. Not much changed since 3.2.0. From my point of view there should not be a problem with that.

mariushoch:
We should IMO tag 3.4.0 here and use that in Wikibase master.
For the build, I'll just check out the patch individually.

Benestar:
Until when does the 3.4 release have to be tagged then?

mariushoch:
> Until when does the 3.4 release have to be tagged then?

ASAP, I want to backport this today, hopefully before SWAT.

PR #120, opened 2016-03-04
Streamline FingerprintPatcher

thiemowmde:
This streamlines the implementation and avoids calling the methods `Fingerprint::setLabels`, `setDescriptions` and `setAliasGroups`.
- Instead of turning the Term/AliasGroupLists into arrays, diffing and patching these arrays, and then turning them back into completely new lists of Term/AliasGroup objects, the patcher now does what it's name suggests: apply the patch. Done.
- This refactoring makes it very easy to turn this patcher into separate Label-, Description- and AliasGroupPatchers when we need them (see https://github.com/wmde/WikibaseDataModel/pull/513 and https://github.com/wmde/WikibaseDataModel/pull/632).

Missing:
- ~~Refactor alias patching the same way.~~ Done.
- ~~Add tests. Lots and lots of them. All the conflicts are currently not tested!~~ Done.
- ~~Split code. Bad nesting.~~ Done.

[Bug: T78298](http://phabricator.wikimedia.org/T78298)

### DIFF
--- a/src/Diff/Internal/FingerprintPatcher.php
+++ b/src/Diff/Internal/FingerprintPatcher.php
@@ -2,9 +2,14 @@
 
 namespace Wikibase\DataModel\Services\Diff\Internal;
 
+use Diff\DiffOp\AtomicDiffOp;
 use Diff\DiffOp\Diff\Diff;
+use Diff\DiffOp\DiffOp;
+use Diff\DiffOp\DiffOpAdd;
+use Diff\DiffOp\DiffOpChange;
+use Diff\DiffOp\DiffOpRemove;
 use Diff\Patcher\MapPatcher;
-use InvalidArgumentException;
+use Diff\Patcher\PatcherException;
 use Wikibase\DataModel\Services\Diff\EntityDiff;
 use Wikibase\DataModel\Term\AliasGroupList;
 use Wikibase\DataModel\Term\Fingerprint;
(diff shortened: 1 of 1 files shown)


thiemowmde:
I did an additional benchmark. Memory usage is unaffected. Runtime drops from 500ms to 200ms in my worst case test.

thiemowmde:
Rebased on top of #123.

brightbyte:
Removing milestone after discussion with Thiemo. This is a non-critical refactoring.

JeroenDeDauw:
It's somewhat unfortunate to have to write custom patching code like this. I can see the benefits, though to me it's unclear if those outweigh the drawbacks. Don't let that stop you though, if you think this is better and worth the time, then go for it.

thiemowmde:
Rebased on top of #124.

src/Diff/Internal/FingerprintPatcher.php
```diff
@@ -41,30 +37,164 @@ public function patchFingerprint( Fingerprint $fingerprint, EntityDiff $patch )
...
+	/**
+	 * @see ListPatcher
+	 *
+	 * @param string[] $aliases
+	 * @param Diff $patch
+	 *
+	 * @throws PatcherException
+	 * @return string[]
+	 */
+	private function getPatchedAliases( array $aliases, Diff $patch ) {
```
thiemowmde, line 175: Note: This whole method could be replaced with a `ListPatcher` call, but this needs https://github.com/wmde/Diff/pull/65.

PR #121, opened 2016-03-09
Add compatibility with DataModel 6.0

Benestar:


JeroenDeDauw:
+1

JeroenDeDauw:
Looks good so far

tests/unit/Diff/EntityDiffOldTest.php
```diff
@@ -127,9 +127,14 @@ public static function generateApplyData( $entityType ) {
...
 	/**
+	 * @return array[]
+	 */
+	public abstract function provideApplyData();
+
+	/**
 	 * @dataProvider provideApplyData
 	 */
-	public function testApply( Entity $a, Entity $b ) {
+	public function testApply( Item $a, Item $b ) {
```
thiemowmde, line 137: Obviously a design issue, but since Item is not introduced in this patch (see the use section), this is a very nice, minimal migration path.

tests/unit/Diff/EntityDiffOldTest.php
```diff
@@ -127,9 +127,14 @@ public static function generateApplyData( $entityType ) {
...
 		$differ = new EntityDiffer();
```
thiemowmde, line 138: This is dead code. testApply can be abstract.

thiemowmde:
A very good, minimal patch. Please do not make it more complicated. I just left some notices for later patches.

PR #124, opened 2016-03-14
Minimize FingerprintPatcher implementation

thiemowmde:
This patch removes redundant code:
- Labels and descriptions use the same method now.
- `Fingerprint::setLabels`, `setDescriptions` and `setAliaseGroups` are not called any more. Instead, the original `TermList`s and `AliasGroupList` are patched. You could argue that this requires DataModel 6.0, because we changed the provider interfaces in 6.0 to guarantee they return objects by reference. However, all previous DataModel versions already did that. The tests show that this change is indeed compatible with the previous DataModel versions.
- `getAliasesArrayForPatching` was a duplicate implementation of `AliasGroupList::toTextArray`.


brightbyte:
Removing milestone after discussion with Thiemo. This is a non-critical refactoring.

brightbyte:
Removing milestone after discussion with Thiemo. This is a non-critical refactoring.

PR #125, opened 2016-03-16
Introduce TermListPatcher and AliasGroupListPatcher

Benestar:
This pull request is mainly about moving code around and creating a lot of test cases.

It consists of three commits that can be reviewed individually.

After this pull request, FingerprintPatcher is unused and can be removed. We have to make sure that all test cases in FingerprintPatcherTest are also present in TermListPatcherTest and AliasGroupListPatcher test.

To make the new classes actually useful, we have to move them out of the Internal namespace and make them package public.

### DIFF
--- a/tests/unit/Diff/Internal/AliasGroupListPatcherTest.php
+++ b/tests/unit/Diff/Internal/AliasGroupListPatcherTest.php
@@ -1,0 +1,250 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Tests\Diff\Internal;
+
+use Diff\DiffOp\Diff\Diff;
+use Diff\DiffOp\DiffOpAdd;
+use Diff\DiffOp\DiffOpChange;
+use Diff\DiffOp\DiffOpRemove;
+use PHPUnit_Framework_TestCase;
+use Wikibase\DataModel\Services\Diff\Internal\AliasGroupListPatcher;
+use Wikibase\DataModel\Term\AliasGroup;
+use Wikibase\DataModel\Term\AliasGroupList;
+
+/**
+ * @covers Wikibase\DataModel\Services\Diff\Internal\AliasGroupListPatcher
+ *
+ * @license GPL-2.0+
+ * @author Bene* < benestar.wikimedia@gmail.com >
+ */
+class AliasGroupListPatcherTest extends PHPUnit_Framework_TestCase {
+
+	/**
+	 * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
+	 */
+	public function providePatchAliasGroupList() {
+		return array(
+			'add aliases (associative)' => array(
+				new AliasGroupList(),
+				new Diff( array(
+					'en' => new Diff( array( new DiffOpAdd( 'foo' ), new DiffOpAdd( 'bar' ) ) )
+				), true ),
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar' ) )
+				) )
+			),
+			'add aliases (non-associative)' => array(
+				new AliasGroupList(),
+				new Diff( array(
+					'en' => new Diff( array( new DiffOpAdd( 'foo' ), new DiffOpAdd( 'bar' ) ) )
+				), false ),
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar' ) )
+				) )
+			),
+			'add aliases (auto-detected)' => array(
+				new AliasGroupList(),
+				new Diff( array(
+					'en' => new Diff( array( new DiffOpAdd( 'foo' ), new DiffOpAdd( 'bar' ) ) )
+				) ),
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar' ) )
+				) )
+			),
+			'add aliases (atomic)' => array(
+				new AliasGroupList(),
+				new Diff( array(
+					'en' => new DiffOpAdd( array( 'foo', 'bar' ) )
+				) ),
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar' ) )
+				) )
+			),
+			'change aliase' => array(
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar' ) )
+				) ),
+				new Diff( array(
+					'en' => new Diff( array( new DiffOpChange( 'bar', 'baz' ) ) )
+				) ),
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'baz' ) )
+				) )
+			),
+			'change aliases (atomic)' => array(
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar' ) )
+				) ),
+				new Diff( array(
+					'en' => new DiffOpChange( array( 'foo', 'bar' ), array( 'baz' ) )
+				) ),
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'baz' ) )
+				) )
+			),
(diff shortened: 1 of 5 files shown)


src/Diff/ItemPatcher.php
```diff
@@ -18,9 +19,14 @@
 class ItemPatcher implements EntityPatcherStrategy {
```
thiemowmde, line 18: Uh, **this** is weird in my opinion. Item and Property are guaranteed to have a Fingerprint. We do not win anything by using the smaller patchers here.

src/Diff/PropertyPatcher.php
```diff
@@ -17,9 +18,14 @@
 class PropertyPatcher implements EntityPatcherStrategy {
```
thiemowmde: As above.

tests/unit/Diff/Internal/AliasGroupListPatcherTest.php
```diff
@@ -0,0 +1,229 @@
...
+ */
+class AliasGroupListPatcherTest extends PHPUnit_Framework_TestCase {
+
+	public function providePatchAliasGroupList() {
+		return array(
+			'add aliases' => array(
+				new AliasGroupList(),
+				new Diff( array(
+					'en' => new Diff( array( new DiffOpAdd( 'foo' ), new DiffOpAdd( 'bar' ) ) )
+				) ),
```
thiemowmde, line 49: I notice there is never a true/false here, so all `Diff` instances rely on their "is associative" auto-detection. I would either avoid that or add tests for all three possible values true/false/null.

Benestar:
I removed the questioned commit from the chain.

thiemowmde:
PHPMD. :-(

tests/unit/Diff/Internal/AliasGroupListPatcherTest.php
```diff
@@ -0,0 +1,229 @@
...
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar' ) )
+				) )
+			),
+			'change aliases is no-op' => array(
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar' ) )
+				) ),
+				new Diff( array(
+					'en' => new Diff( array( new DiffOpChange( 'bar', 'baz' ) ) )
```
thiemowmde, line 68: This looks like a valid change. Why should this be a no-op? Especially since the very similar change below works.

tests/unit/Diff/Internal/AliasGroupListPatcherTest.php
```diff
@@ -0,0 +1,229 @@
...
+					new AliasGroup( 'en', array( 'foo', 'bar', 'baz' ) )
+				) ),
+				new Diff( array(
+					'en' => new DiffOpRemove( array( 'foo', 'bar' ) )
+				) ),
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar', 'baz' ) )
+				) )
+			),
+			'add an existing language is no-op' => array(
```
thiemowmde: The test is correct, but the description should not talk about a "no-op".

tests/unit/Diff/Internal/TermListPatcherTest.php
```diff
@@ -0,0 +1,142 @@
...
+					new Term( 'en', 'foo' ),
+					new Term( 'de', 'bar' ),
+					new Term( 'nl', 'baz' )
+				) ),
+				new Diff( array(
+					'en' => new DiffOpChange( 'foo', 'bar' ),
+					'de' => new DiffOpRemove( 'bar' ),
+					'nl' => new DiffOpChange( 'bar', 'foo' ),
+					'it' => new DiffOpAdd( 'foo' )
+				) ),
```
thiemowmde, line 161: All `Diff` instances that contain language keys need to be marked as associative with `true`.

tests/unit/Diff/Internal/AliasGroupListPatcherTest.php
```diff
@@ -0,0 +1,229 @@
...
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo', 'bar' ) )
+				) )
+			),
+			'change different aliases is no-op' => array(
+				new AliasGroupList( array(
+					new AliasGroup( 'en', array( 'foo' ) )
+				) ),
+				new Diff( array(
+					'en' => new Diff( array( new DiffOpChange( 'bar', 'baz' ) ) )
```
thiemowmde, line 202: All test cases that are nested like this (`Diff` → `Diff` → `DiffOpChange`) should be duplicated, one with `Diff` → `Diff( …, true )` → `DiffOpChange` and one with `Diff` → `Diff( …, false )` → `DiffOpChange`.

tests/unit/Diff/Internal/AliasGroupListPatcherTest.php
Benestar: C&p mistake, will fix

tests/unit/Diff/Internal/AliasGroupListPatcherTest.php
Benestar, line 68: If I understand it correctly, that is basically what https://github.com/wmde/Diff/pull/65 is about. I just wanted to test the current behaviour but we should change this test if the other pull request gets merged.

tests/unit/Diff/Internal/TermListPatcherTest.php
Benestar, line 161: How does this still work without explicitely stating that it is associative?

Benestar:
> I notice there is never a true/false here, so all Diff instances rely on their "is associative" auto-detection. I would either avoid that or add tests for all three possible values true/false/null.

I didn't yet completely understand what the $isAssociative is about. Is it just the array of diffs that can be associative or linear? Anyways, it would perhaps be nice to do that in the actual test so that we don't have too much duplication.

> PHPMD. :-(

Any idea how to fix that? Is it possible to disable the linting per file/method?

tests/unit/Diff/Internal/TermListPatcherTest.php
JeroenDeDauw, line 161: When not stated explicitly, it is determined by seeing if there are changed that are no add or remove.

JeroenDeDauw:
> I didn't yet completely understand what the $isAssociative is about. Is it just the array of diffs that can be associative or linear?

The diffs as well. See the first 3 sections in "usage" at https://github.com/wmde/Diff#usage

JeroenDeDauw:
> To make the new classes actually useful, we have to move them out of the Internal namespace and make them package public.

Sounds good. I'm guessing this was not done so far as we actually had no need to expose those things. I don't actually recall the reason though.

thiemowmde:
Rebased. Note that this patch, as it is now, reverts #120! I suggest we fix #131 first before continuing to work on this.

tests/unit/Diff/Internal/AliasGroupListPatcherTest.php
thiemowmde, line 68: Fixed in #131.

thiemowmde:
Rebased again, on top of #120 and #131. All this patch now does is splitting the existing implementation (now without reverting anything) and adding tests. The only relevant detail I had to change in the tests was the one #131 fixed (see https://github.com/wmde/WikibaseDataModelServices/pull/125#discussion_r61218037). I also added a few more test cases with the "associative" flag set to true and false.

From my point of view this is ready to be merged.

Open questions (for other patches, should not block this one):
- Which classes must be public?
- Can we fix and use https://github.com/wmde/Diff/pull/65 before releasing this?

adrianheine:
Needs manual rebase

PR #128, opened 2016-03-24
Add missing type hints to getters that return mocks

thiemowmde:
I found candidates with a pretty complicated regex, searching for functions that construct and return a mock but do not have a PHPDoc block. This improves type safety and reduces the number of warnings analysis tools like PHPStorm report.

### DIFF
--- a/tests/unit/EntityId/EntityIdLabelFormatterTest.php
+++ b/tests/unit/EntityId/EntityIdLabelFormatterTest.php
@@ -7,6 +7,7 @@ use Wikibase\DataModel\Entity\EntityId;
 use Wikibase\DataModel\Entity\ItemId;
 use Wikibase\DataModel\Entity\PropertyId;
 use Wikibase\DataModel\Services\EntityId\EntityIdLabelFormatter;
+use Wikibase\DataModel\Services\Lookup\LabelDescriptionLookup;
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Such PHP5

thiemowmde:
The need for these type hints is the same in PHP7.

JeroenDeDauw:
You don't need the docs in PHP7

thiemowmde:
Even in the latest PHPUnit version `getMock` returns `PHPUnit_Framework_MockObject_MockObject`, see https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php#L1406. With this the IDE can not know what the exact type of the returned mock is. I don't see how this is different with PHP7.

PS: https://github.com/sebastianbergmann/phpunit-mock-objects/blob/master/src/Framework/MockObject/Generator.php#L138

JeroenDeDauw:
Such Thiemo

thiemowmde:
That was meant as a question. How is this different with PHP7?

JeroenDeDauw:
> You don't need the docs in PHP7

... because it has return type hints

thiemowmde:
So PHP7 **does** need type hints. Thanks.

PR #130, opened 2016-04-20
Disable email spam from Travis

thiemowmde:
I guess the Travis guys have changed the default, which I find very annoying.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ install: travis_retry composer require "wikibase/data-model=$DM" --prefer-source
 script: composer ci
 
 notifications:
+  email: false
   irc:
     channels:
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Default was not changed AFAIK. This was on before. You only get mails when you break the build or fix it.

How about disabling it for PRs/branches? There we have CR anyway. I'd like to keep the emails when master gets broken.

thiemowmde:
I'm not sure what else I could do here. Other Travis configurations in other components contain your email. Feel free to do the same here, but please disable it for all other users that did not opted-in.

All I know is that I never got these emails before today, and I do not know what changed.

thiemowmde:
For now I "logged out" of Travis by revoking it's right to use my GitHub login.

PR #133, opened 2016-04-25
Use more readable switch statements in FingerprintPatcher

thiemowmde:
Pure refactoring for readability, as suggested by @Jonas-WMDE. Instead of a series of `instanceof`s we use `DiffOp::getType()` now.

[Bug: T78298](http://phabricator.wikimedia.org/T78298)

### DIFF
--- a/src/Diff/Internal/FingerprintPatcher.php
+++ b/src/Diff/Internal/FingerprintPatcher.php
@@ -61,24 +61,34 @@ class FingerprintPatcher {
 	private function patchTerm( TermList $terms, $lang, AtomicDiffOp $diffOp ) {
 		$hasLang = $terms->hasTermForLanguage( $lang );
 
-		if ( $diffOp instanceof DiffOpAdd ) {
-			if ( !$hasLang ) {
-				$terms->setTextForLanguage( $lang, $diffOp->getNewValue() );
-			}
-		} elseif ( $diffOp instanceof DiffOpChange ) {
-			if ( $hasLang
-				&& $terms->getByLanguage( $lang )->getText() === $diffOp->getOldValue()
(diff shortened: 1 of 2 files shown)


src/Diff/Internal/FingerprintPatcher.php
```diff
@@ -61,24 +61,34 @@ private function patchTermList( TermList $terms, Diff $patch ) {
...
-		} elseif ( $diffOp instanceof DiffOpRemove ) {
-			if ( $hasLang
-				&& $terms->getByLanguage( $lang )->getText() === $diffOp->getOldValue()
-			) {
-				$terms->removeByLanguage( $lang );
-			}
-		} else {
-			throw new PatcherException( 'Invalid terms diff' );
+		switch ( $diffOp->getType() ) {
+			case 'add':
```
JeroenDeDauw: You can still do the `instanceof` when switching on `true`

thiemowmde: Uh, thanks for the pointer. I remember seeing this trick somewhere. I will give it a try.

thiemowmde: Micro-benchmarked (in PHP 5.5) out of curiosity:

``` php
if ( $element->getType() === 'type' ) {
    … // 700 ms

switch ( $element->getType() ) {
    case 'type':
        … // 480 ms

switch ( get_class( $element ) ) {
    case Class::class:
        … // 420 ms

switch ( true ) {
    case $element instanceof Class:
        … // 220 ms

if ( $element instanceof Class ) {
    … // 160 ms
```

JeroenDeDauw:
+1

PR #136, opened 2016-04-27
Add missing Item and PropertyPatcher tests

thiemowmde:
As discussed in https://gerrit.wikimedia.org/r/285163.


thiemowmde:
I'm fine with releasing 3.6 without this, but what would be the reason to not simply merge this? Pinging @JonasKress, @adrianheine and @brightbyte since he was asking for this, if I remember correctly.

thiemowmde:
Rebased.

brightbyte:
Sorry, I didn't see this patch before.

PR #137, opened 2016-04-28
Rename Suffix… to PrefixedEntityIdParser

thiemowmde:
This is not a parser for entity IDs with suffixes. We (@adrianheine and I) found this confusing today.

This is used in two places in the top level factories of Wikibase.git, nowhere else.

### DIFF
--- a/src/EntityId/PrefixedEntityIdParser.php
+++ b/src/EntityId/PrefixedEntityIdParser.php
@@ -15,7 +15,7 @@ use Wikibase\DataModel\Entity\EntityIdParsingException;
  * @license GPL-2.0+
  * @author Daniel Kinzler
  */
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
+2 CR; suggest to wait with merge till concrete plans to do the next breaking release

thiemowmde:
I agree.

brightbyte:
Needs rebase. Is this still wanted/needed?

thiemowmde:
Rebased. I believe this should still be done. The current name sounds like it is a parser that extracts suffixes, but it does the opposite. It parses EntityIDs that are prefixed.

JeroenDeDauw:
Needs rebase again

JeroenDeDauw (approved):

PR #138, opened 2016-05-03
Use differ class name in differStrategies array as key

JonasKress:
[Bug: T133881](https://phabricator.wikimedia.org/T133881)

### DIFF
--- a/src/Diff/EntityDiffer.php
+++ b/src/Diff/EntityDiffer.php
@@ -25,7 +25,8 @@ class EntityDiffer {
 	}
 
 	public function registerEntityDifferStrategy( EntityDifferStrategy $differStrategy ) {
-		$this->differStrategies[] = $differStrategy;
+		$differId = get_class( $differStrategy );
+		$this->differStrategies[$differId] = $differStrategy;
 	}
 
 	/**


adrianheine:
Mh, this would currently work, but there might be different instances of the same class that diff different things.

thiemowmde:
- Misses tests.
- Is this relevant for a Phabricator ticket?
- What Adrian said. The class name is not a good identifier.
- I wonder why this is needed? Sure, to avoid duplicates in the array. But even if there are duplicates, when and how is this a problem? The loop stops the moment it finds the first differ that can diff the given entity type.
- The only effect of this patch is that the differ added last will be picked, instead of the first.
- This makes this a breaking change.
- See #61.

JonasKress:
https://gerrit.wikimedia.org/r/285938

thiemowmde:
In my opinion, a proper fix would be to remove the default constructor. Or even better, deprecate this class and introduce a `DispatchingEntityDiffer`, either here or in repo where it is needed.

In any case, not critical because the duplicates don't do anything.

JeroenDeDauw:
There is no need to make any breaking changes to this class. I've looked at the commit on Gerrit and can see how what this class provides is not ideal for that user. What I suggest is done is to extract the relevant code from this class into a new one that requires the strategies in its constructor and has no default ones. Existing users remain well served and unaffected, and the Gerrit one can use the new class.

JeroenDeDauw:
Presumably no longer relevant. Please reopen if wrong

PR #140, opened 2016-05-06
Update DataModel version matrix in .travis.yml

thiemowmde:
- PHP 5.6 was tested twice but PHP 5.5 was not.
- Testing with 4.3@dev is obsolete with 5.0.

Same in:
- https://github.com/wmde/WikibaseDataModelSerialization/pull/202
- https://github.com/wmde/WikibaseInternalSerialization/pull/108

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,18 +6,18 @@ env:
 matrix:
   fast_finish: true
   include:
-    - env: DM=~4.2
+    - env: DM=~6.0
       php: 5.3
-    - env: DM=~4.2
+    - env: DM=@dev
       php: 5.4
-    - env: DM=~4.3@dev
+    - env: DM=~6.0
+      php: 5.5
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,18 +6,18 @@ env:
 matrix:
   fast_finish: true
   include:
-    - env: DM=~4.2
+    - env: DM=~6.0
       php: 5.3
-    - env: DM=~4.2
+    - env: DM=@dev
```
JeroenDeDauw, line 11: I'd have put the @dev with a latest PHP version, so this does not break when changing DM master to require PHP 5.5+

thiemowmde, line 11: Hm, ok. Simply moving the `@dev` up to PHP 5.5 should be ok. Is it ok when I do this edit without a pull request?

PR #135, opened 2016-04-25
Introduce StatementListPatcher::patchStatementList

thiemowmde:
Needed in https://gerrit.wikimedia.org/r/#/c/285163/3/src/DataModel/Services/Diff/MediaInfoPatcher.php.

[Bug: T78298](http://phabricator.wikimedia.org/T78298)
[Bug: T132442](http://phabricator.wikimedia.org/T132442)

PR #141, opened 2016-05-06
Fix critical StatementListPatcher regressions

thiemowmde:
#135 introduced several regressions that had no tests:
- Adding a statement with the same GUID should be ignored and not create duplicates.
- Changing a non-existing statement should be ignored and not add the new statement without removing anything.
- Changing a statement should not change it's index position in the array. This is pretty critical because this will be visible in our database dumps.

The helper method I'm introducing here will become a lot simpler with https://github.com/wmde/WikibaseDataModel/pull/664.

I wrote these tests with the previous 3.5 release and made sure they succeed in both versions. You may want to verify that.

### DIFF
--- a/src/Diff/Internal/StatementListPatcher.php
+++ b/src/Diff/Internal/StatementListPatcher.php
@@ -40,15 +40,21 @@ class StatementListPatcher {
 			switch ( true ) {
 				case $diffOp instanceof DiffOpAdd:
 					/** @var DiffOpAdd $diffOp */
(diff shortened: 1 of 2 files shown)


thiemowmde:
Steps to reproduce (in the UI):
1. Assigning the same GUID is not possible in the UI. This can only be tested in a unit test.
2. Open two tabs with the same item. In both tabs start editing the same statement. Hit delete in one tab and save in the other. There should be an edit conflict.
3. It's possible that the index position bug can be seen in …/Special:EntityData/….json.

PR #142, opened 2016-05-09
Make differs/patchers required by MediaInfo public

thiemowmde:
Note that this conflicts with #141! Please merge #141 first.

For MediaInfo this is a breaking change because this will break the current implementation that uses classes that are in the "Internal" namespace and marked as "packed private" in the class level documentation. However, from the perspective of this component this is not a breaking change and should, in my opinion, go into the planned 3.6 release.


JeroenDeDauw:
+1

JonasKress:
Needs rebase

thiemowmde:
Rebased.

JeroenDeDauw:
Tests went boom

thiemowmde:
A test introduced in #141 broke after rebase. Fixed now.

PR #145, opened 2016-07-21
Make use of DataModel 6.1 features

thiemowmde:
~~Please do not merge to early. Wait for other reasons to have a breaking 4.0 release please.~~


JeroenDeDauw:
Breaking how?

thiemowmde:
Dropping compatibility with older versions is something I always consider some kind of "breaking" change. It's none, you are right.

JeroenDeDauw:
I was not making a point, just asking a question. So I can't be right ;p

brightbyte:
Please rebase

thiemowmde:
Rebased.

PR #148, opened 2016-09-22
Raise PHP version requirement to 5.5

manicki:
Following the change in data model: https://github.com/wmde/WikibaseDataModel/pull/680

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,15 +6,13 @@ env:
 matrix:
   fast_finish: true
   include:
-    - env: DM=~6.0
-      php: 5.3
-    - env: DM=~6.0
-      php: 5.4
     - env: DM=@dev
       php: 5.5
-    - env: DM=~5.0
+    - env: DM=~6.0
+      php: 5.5
+    - env: DM=~6.0
       php: 5.6
-    - env: DM=~4.2
+    - env: DM=~5.0
       php: 7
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,10 +6,6 @@ env:
 matrix:
   fast_finish: true
   include:
-    - env: DM=~6.0
-      php: 5.3
-    - env: DM=~6.0
```
thiemowmde, line 11: You are removing all tests with DataModel 6 here. You need to move these version numbers down, at least one level, so every version is at least tested once.

manicki: So now I moved those version number down (if I understood you correctly). I've also added an extra pair DM 6.0, PHP 5.5. to have a PHP 5.5 build that is not allowed to fail.
I don't know if such multiple builds per language version are something that is wanted here, though.

thiemowmde: When I worked on this matrix, my idea was to list the PHP versions in increasing order, and DM versions in decreasing order, possibly maximizing compatibility.

```
    - env: DM=@dev
      php: 5.5
    - env: DM=~6.0
      php: 5.6
    - env: DM=~5.0
      php: 7
    - env: DM=~4.2
      php: hhvm
```

What you have now is fine. Thanks.

PR #149, opened 2016-09-29
Add PrefixMappingEntityIdParserFactory

manicki:
Note: this contains commit(s) from https://github.com/wmde/WikibaseDataModelServices/pull/146. ~~CI tests will be failing as longs as that PR's tests fail (ie. until DataModel 6.2 is tagged).~~

The actual change of this PR is at https://github.com/wmde/WikibaseDataModelServices/pull/149/commits/35cb120b77c6a309e30245baec8d14459c9347a7.

Per what we discussed in https://github.com/wmde/WikibaseDataModelServices/pull/146 I will be making any improvements to this PR by adding commits on top of the first one, and finally squash all of them once the change is ready to go.

Ticket: [T146274](https://phabricator.wikimedia.org/T146274)

### DIFF
--- a/src/EntityId/PrefixMappingEntityIdParserFactory.php
+++ b/src/EntityId/PrefixMappingEntityIdParserFactory.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Wikibase\DataModel\Services\EntityId;
+
+use Wikibase\DataModel\Entity\EntityIdParser;
+use Wikimedia\Assert\Assert;
+use Wikimedia\Assert\ParameterAssertionException;
+
+/**
+ * @since 3.7
+ *
+ * @license GPL-2.0+
+ */
+class PrefixMappingEntityIdParserFactory {
+
+	/**
+	 * @var string[]
+	 */
+	private $idPrefixMapping;
+
+	/**
+	 * @var EntityIdParser
+	 */
+	private $parser;
+
+	/**
+	 * @param EntityIdParser $parser
+	 * @param array $idPrefixMapping An associative array mapping repository names (strings) to id serialization
+	 *        prefix mappings specific to the particular repository (@see PrefixMappingEntityIdParser).
+	 *        If an empty-string key is provided in the mapping for some repository, its value must be the same
+	 *        as the repository name.
+	 *
+	 * @throws ParameterAssertionException
+	 */
+	public function __construct( EntityIdParser $parser, array $idPrefixMapping ) {
+		Assert::parameterElementType( 'string', array_keys( $idPrefixMapping ), 'array_keys( $idPrefixMapping)' );
+		foreach ( $idPrefixMapping as $repositoryName => $mapping ) {
+			Assert::parameter(
+				strpos( $repositoryName, ':' ) === false,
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -23,7 +23,8 @@
 		"php": ">=5.5.0",
 		"wikibase/data-model": "~6.0|~5.0|~4.2",
 		"data-values/data-values": "~0.1|~1.0",
-		"diff/diff": "~2.0|~1.0"
+		"diff/diff": "~2.0|~1.0",
```
JeroenDeDauw, line 26: Also need to bump DataModel version

manicki: True. Did it in https://github.com/wmde/WikibaseDataModelServices/pull/146, we still need to get DM 6.2 though

brightbyte:
I left a comment on 7aa8cbaca43d. Except for the issue mentioned there, the code looks good.

brightbyte:
I replied at 7aa8cba. Perhaps we want to change the constructor of PrefixMappingEntityIdParser.

manicki:
I've adjusted thus pull requested after changes in https://github.com/wmde/WikibaseDataModelServices/pull/146 (per what has been discussed here at https://github.com/wmde/WikibaseDataModelServices/commit/7aa8cbaca43dcff431b2eb319ff67fb24cd04034).

Due to github being bound to branches it is difficult easily track changes in this PR. If you want to see what has been changed you could try running `git diff -r 3e3e7b5 35cb120` in your local copy of this repo. Just look for changes in PrefixMappingEntityIdParserFactory and PrefixMappingEntityIdParserFactoryTest. There are plenty of noise in this diff due to rebase (switch to short array syntax etc).

One controversial thing here that the factory constructor expects a prefix mapping array in a form `[ 'foo' => $fooMappings ]`. $fooMappings will be passed to PrefixMappingEntityIdParser constructor. Now the controversial part: $fooMappings might but is not required to contain a default mapping (`'' => 'foo'`). If the empty string key is not present in the mapping for repository 'foo', factory adds a `'' => 'foo'` pair and creates a parser instance. If such mapping is provided to the factory but it is not consistent with the repository name (ie. `[ 'foo' => [ '' => 'bar' ] ]`) the factory will throw an exception.

It might be more clear to always require empty-string mapping to be specified explicitly. I haven't done that as I thought that the array passed to the factory would be coming from some config setting where one possibly doesn't want to have redundant `'' => 'foo'` mappings for repository 'foo'.

Also, the factory currently allows to have no mappings defined for the repo (see https://github.com/wmde/WikibaseDataModelServices/pull/149/commits/35cb120b77c6a309e30245baec8d14459c9347a7#diff-c3904bfe92576d2c6aba86ecaa2a0083R34), resulting with the parser instance simply prefixing all entity ids with the repository name being returned by the factory.

I would be happy to change those possibly not clear assumptions and make the factory expect all things defined explicitly.

addshore:
My 2p on this patch is that both classes and methods should have @since tags ideally!
Other than that it all looks good, and i'll leave it to @brightbyte

brightbyte:
@addshore: class level seems sufficient for me.

PR #150, opened 2016-10-06
Add DispatchingEntityLookup

manicki:
DispatchingEntityLookup picks the right EntityLookup object to perform the lookup depending on the repository the input EntityId object is coming from.

Issues to be solved:
- how do we want to name such services (this is the first of the whole bunch): should they be "Multiplexing" services, or rather "Dispatching" services, or maybe something else? Any suggestions? - edit 12.10.2016: per Jeroen's suggestion "Dispatching" is used in the service name, not the original "Multiplexing"
- ~~depends on DM 6.2 which has not been tagged yet,~~
- ~~I am not quite sure how it should deal with the "main" repository, ie. how should it handle "local" EntityId (ie. not coming from the foreign repository). Should the constructor get a single "default" EntityLookup for all non-foreign EntityIds, and then a (possibly empty) list of repository speciific lookups? Or maybe rather should it expect to only get an array of lookups with special empty-string key reserved for the default lookup? The latter seems far more hacky, so I guess it is either the former approach or something even better I couldn't think of. Suggestions?~~
- to make review easier amendments to the original commit have been submitted as separate commits. They all need to be squashed into single commit before merging.

~~Note: this includes https://github.com/wmde/WikibaseDataModelServices/pull/150/commits/336a8341e7db350ec72b31696ea845167260daa8 from https://github.com/wmde/WikibaseDataModelServices/pull/154~~

Ticket: [T146989](https://phabricator.wikimedia.org/T146989)

### DIFF
--- a/src/Lookup/DispatchingEntityLookup.php
+++ b/src/Lookup/DispatchingEntityLookup.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Lookup;
+
+use Wikibase\DataModel\Assert\RepositoryNameAssert;
+use Wikibase\DataModel\Entity\EntityDocument;
+use Wikibase\DataModel\Entity\EntityId;
+use Wikimedia\Assert\Assert;
+use Wikimedia\Assert\ParameterAssertionException;
+
+/**
+ * Delegates lookup to the repository-specific EntityLookup
+ * based on the name of the repository an EntityId belongs to.
+ * This class does not strip repository prefixes of incoming
+ * entity IDs.
+ *
+ * @since 3.7
+ *
+ * @license GPL-2.0+
+ */
+class DispatchingEntityLookup implements EntityLookup {
+
+	/**
+	 * @var EntityLookup[]
+	 */
+	private $lookups;
+
+	/**
+	 * @since 3.7
+	 *
+	 * @param EntityLookup[] $lookups associative array with repository names (strings) as keys
+	 *                                and EntityLookup objects as values.
+	 *
+	 * @throws ParameterAssertionException
+	 */
+	public function __construct( array $lookups ) {
+		Assert::parameter(
+			!empty( $lookups ),
+			'$lookups',
+			'must must not be empty'
+		);
+		Assert::parameterElementType( EntityLookup::class, $lookups, '$lookups' );
+		RepositoryNameAssert::assertParameterKeysAreValidRepositoryNames( $lookups, '$lookups' );
+		$this->lookups = $lookups;
+	}
+
+	/**
+	 * @see EntityLookup::getEntity
+	 *
+	 * @since 3.7
+	 *
+	 * @param EntityId $entityId
+	 *
+	 * @return null|EntityDocument
+	 * @throws EntityLookupException
+	 * @throws UnknownForeignRepositoryException
+	 */
+	public function getEntity( EntityId $entityId ) {
+		$lookup = $this->getLookupForEntityId( $entityId );
+		return $lookup->getEntity( $entityId );
+	}
+
+	/**
+	 * @see EntityLookup::hasEntity
+	 *
+	 * @since 3.7
+	 *
+	 * @param EntityId $entityId
+	 *
+	 * @return bool
+	 * @throws EntityLookupException
+	 * @throws UnknownForeignRepositoryException
+	 */
+	public function hasEntity( EntityId $entityId ) {
+		$lookup = $this->getLookupForEntityId( $entityId );
+		return $lookup->hasEntity( $entityId );
+	}
+
+	/**
+	 * @param EntityId $entityId
+	 *
+	 * @return EntityLookup
+	 * @throws UnknownForeignRepositoryException
+	 */
+	private function getLookupForEntityId( EntityId $entityId ) {
+		$repo = $entityId->getRepositoryName();
+		if ( !isset( $this->lookups[$repo] ) ) {
+			throw new UnknownForeignRepositoryException( $entityId->getRepositoryName() );
+		}
+		return $this->lookups[$repo];
+	}
+
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,13 @@ matrix:
   include:
     - env: DM=@dev
       php: 5.5
-    - env: DM=~6.2
+    - env: DM=~6.3
       php: 5.5
-    - env: DM=~6.2
+    - env: DM=~6.3
       php: 5.6
-    - env: DM=~6.2
+    - env: DM=~6.3
       php: 7
-    - env: DM=~6.2
+    - env: DM=~6.3
       php: hhvm
   exclude:
     - env: THENEEDFORTHIS=FAIL
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
 	},
 	"require": {
 		"php": ">=5.5.0",
-		"wikibase/data-model": "~6.2",
+		"wikibase/data-model": "~6.3",
 		"data-values/data-values": "~0.1|~1.0",
 		"diff/diff": "~2.0|~1.0",
 		"wikimedia/assert": "~0.2.2"
--- a/tests/unit/Lookup/DispatchingEntityLookupTest.php
+++ b/tests/unit/Lookup/DispatchingEntityLookupTest.php
@@ -1,0 +1,141 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Tests\Lookup;
+
+use Exception;
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Services\Fixtures\FakeEntityDocument;
+use Wikibase\DataModel\Services\Lookup\DispatchingEntityLookup;
+use Wikibase\DataModel\Services\Lookup\EntityLookup;
+use Wikibase\DataModel\Services\Lookup\EntityLookupException;
+use Wikibase\DataModel\Services\Lookup\InMemoryEntityLookup;
+use Wikibase\DataModel\Services\Lookup\UnknownForeignRepositoryException;
+use Wikimedia\Assert\ParameterAssertionException;
+
+/**
+ * @covers Wikibase\DataModel\Services\Lookup\DispatchingEntityLookup
+ *
+ * @license GPL-2.0+
+ */
+class DispatchingEntityLookupTest extends \PHPUnit_Framework_TestCase {
+
+	/**
+	 * @dataProvider provideInvalidForeignLookups
+	 */
+	public function testGivenInvalidForeignLookups_exceptionIsThrown( array $lookups ) {
+		$this->setExpectedException( ParameterAssertionException::class );
+		new DispatchingEntityLookup( $lookups );
+	}
+
+	public function provideInvalidForeignLookups() {
+		return [
+			'no lookups given' => [ [] ],
+			'not an implementation of EntityLookup given as a lookup' => [
+				[ '' => new ItemId( 'Q123' ) ],
+			],
+			'non-string keys' => [
+				[ '' => new InMemoryEntityLookup(), 100 => new InMemoryEntityLookup(), ],
+			],
+			'repo name containing colon' => [
+				[ '' => new InMemoryEntityLookup(),	'fo:oo' => new InMemoryEntityLookup(), ],
+			],
+		];
+	}
+
+	public function testGivenExistingEntityId_getEntityReturnsTheEntity() {
+		$localLookup = new InMemoryEntityLookup();
+		$localLookup->addEntity( new FakeEntityDocument( new ItemId( 'Q1' ) ) );
+		$fooLookup = new InMemoryEntityLookup();
+		$fooLookup->addEntity( new FakeEntityDocument( new PropertyId( 'foo:P11' ) ) );
+
+		$dispatchingLookup = new DispatchingEntityLookup( [ '' => $localLookup, 'foo' => $fooLookup ] );
+
+		$expected = new FakeEntityDocument( new ItemId( 'Q1' ) );
+		$actual = $dispatchingLookup->getEntity( new ItemId( 'Q1' ) );
+		$this->assertTrue( $actual->equals( $expected ) );
+		$this->assertTrue( $actual->getId()->equals( new ItemId( 'Q1' ) ) );
+
+		$expected = new FakeEntityDocument( new PropertyId( 'foo:P11' ) );
+		$actual = $dispatchingLookup->getEntity( new PropertyId( 'foo:P11' ) );
+		$this->assertTrue( $actual->equals( $expected ) );
+		$this->assertTrue( $actual->getId()->equals( new PropertyId( 'foo:P11' ) ) );
+	}
+
+	public function testGivenNotExistingEntityIdFromKnownRepository_getEntityReturnsNull() {
+		$localLookup = new InMemoryEntityLookup();
+		$fooLookup = new InMemoryEntityLookup();
+		$dispatchingLookup = new DispatchingEntityLookup( [ '' => $localLookup, 'foo' => $fooLookup ] );
+		$this->assertNull( $dispatchingLookup->getEntity( new ItemId( 'Q1' ) ) );
+		$this->assertNull( $dispatchingLookup->getEntity( new ItemId( 'foo:Q19' ) ) );
+	}
+
+	/**
+	 * @param Exception $exception
+	 * @return \PHPUnit_Framework_MockObject_MockObject|EntityLookup
+	 */
+	private function getExceptionThrowingLookup( Exception $exception ) {
+		$lookup = $this->getMock( EntityLookup::class );
+		$lookup->expects( $this->any() )
+			->method( $this->anything() )
+			->willThrowException( $exception );
+		return $lookup;
+	}
+
+	public function testLookupExceptionsAreNotCaughtInGetEntity() {
+		$lookup = $this->getExceptionThrowingLookup( new EntityLookupException( new ItemId( 'Q321' ) ) );
+
+		$dispatchingLookup = new DispatchingEntityLookup( [ '' => $lookup ] );
+
+		$this->setExpectedException( EntityLookupException::class );
+		$dispatchingLookup->getEntity( new ItemId( 'Q321' ) );
+	}
+
+	public function testGivenEntityIdFromUnknownRepository_getEntityThrowsException() {
+		$dispatchingLookup = new DispatchingEntityLookup( [ '' => $this->getMock( EntityLookup::class ), ] );
+
+		$this->setExpectedException( UnknownForeignRepositoryException::class );
+
+		$dispatchingLookup->getEntity( new ItemId( 'foo:Q1' ) );
+	}
+
+	public function testGivenExistingEntityId_hasEntityReturnsTrue() {
+		$localLookup = new InMemoryEntityLookup();
+		$localLookup->addEntity( new FakeEntityDocument( new ItemId( 'Q1' ) ) );
+		$fooLookup = new InMemoryEntityLookup();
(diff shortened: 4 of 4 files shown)


src/Lookup/MultiplexingEntityLookup.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+class MultiplexingEntityLookup implements EntityLookup {
+
+	/**
+	 * @var EntityLookup[]
+	 */
+	private $lookups;
+
+	/**
+	 * @param EntityLookup[] $lookups associative array with repository names (strings) as strings
```
jakobw: I think that's supposed to say "as keys"?

src/Lookup/MultiplexingEntityLookup.php
```diff
@@ -0,0 +1,74 @@
...
+	public function getEntity( EntityId $entityId ) {
+		$repo = $entityId->getRepositoryName();
+		if ( !isset( $this->lookups[$repo] ) ) {
+			throw new EntityLookupException( $entityId, 'Unknown repository: ' . $repo );
+		}
+		return $this->lookups[$repo]->getEntity( $entityId );
+	}
+
+	/**
+	 * Returns whether the given entity can bee looked up using
```
jakobw: > bee

:bee:?

src/Lookup/MultiplexingEntityLookup.php
```diff
@@ -0,0 +1,74 @@
...
+	 *
+	 * @return null|EntityDocument
+	 * @throws EntityLookupException
+	 */
+	public function getEntity( EntityId $entityId ) {
+		$repo = $entityId->getRepositoryName();
+		if ( !isset( $this->lookups[$repo] ) ) {
+			throw new EntityLookupException( $entityId, 'Unknown repository: ' . $repo );
+		}
+		return $this->lookups[$repo]->getEntity( $entityId );
```
jakobw: Might be a matter of personal preference but I think it would read better with a private method that takes care of picking the correct repository which can be used here and in `hasEntity` below.

jakobw (commented):
Looks like a solid approach to me!

I like the idea of having the local lookup injected as a constructor argument and then asking the EntityId whether it's local or not before using the repo name for multiplexing.

manicki:
Thanks! I've changed it to have local and foreign lookups separated. It makes sense for me (at least now :). Also did other things you suggested.
For the record: Based from the experience from my previous PRs I've submitted changes as separate commits. They will be squashed into single one when the merge time comes.

jakobw (approved):
I think the public methods probably still need `@since` tags, other than that it looks good to me!

manicki:
Actually, the new class should have @since (and @license) docs. Somehow I managed to not include them here. Fixed now, thanks!

src/Lookup/MultiplexingEntityLookup.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @throws EntityLookupException
+	 */
+	public function hasEntity( EntityId $entityId ) {
+		$lookup = $this->getLookupForEntityId( $entityId );
+		return $lookup->hasEntity( $entityId );
+	}
+
+	/**
+	 * @param EntityId $entityId
+	 * @return EntityLookup
```
jakobw: This method and the ones using it need the `@throws` tag

jakobw (commented):
One small thing missing but the rest looks good to me!

src/Lookup/MultiplexingEntityLookup.php
```diff
@@ -0,0 +1,90 @@
...
+	 */
+	private $lookups;
+
+	/**
+	 * @param EntityLookup[] $lookups associative array with repository names (strings) as keys
+	 *                                and EntityLookup objects as values. Empty-string key
+	 *                                defines lookup for the local repository.
+	 *
+	 * @throws ParameterAssertionException
+	 * @throws ParameterElementTypeException
```
JeroenDeDauw: It's not so nice to force clients to deal with both as they are caused by very similar scenarios. Do they have a common supertype that can be used here instead?

src/Lookup/MultiplexingEntityLookup.php
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Wrapper around repository-specific EntityLookups
+ * picking up the right lookup object for the particular input.
+ *
+ * @since 3.7
+ *
+ * @license GPL-2.0+
+ */
+class MultiplexingEntityLookup implements EntityLookup {
```
JeroenDeDauw: Since I was fuzzy on what exactly multiplexing means, I had to look it up (not good). This service is not combining any of its input, so I think the name is misleading.

We have similar services where we used something like `DispatchingFooService` or `DelegatingFooService`. Would be nice to include a hint that this is done based on repository name.

tests/unit/Lookup/MultiplexingEntityLookupTest.php
```diff
@@ -0,0 +1,165 @@
...
+		$multiplexingLookup = $this->getMultiplexingEntityLookupWithForeignLookups( array( 'foo' => $fooLookup, ) );
+
+		$this->assertFalse( $multiplexingLookup->hasEntity( new ItemId( 'foo:Q3000' ) ) );
+	}
+
+	public function testLookupExceptionsAreNotCaughtInHasEntity() {
+		$lookup = $this->getMock( EntityLookup::class );
+		$lookup->expects( $this->any() )
+			->method( 'hasEntity' )
+			->willThrowException( new EntityLookupException( new ItemId( 'Q321' ) ) );
```
JeroenDeDauw: Happy to see you using `$this->any()`. What about the method part though - is that needed or can you use `$this->anything()`?

tests/unit/Lookup/MultiplexingEntityLookupTest.php
```diff
@@ -0,0 +1,165 @@
...
+	/**
+	 * @dataProvider provideInvalidForeignLookups
+	 */
+	public function testGivenInvalidForeignLookups_exceptionIsThrown( array $lookups ) {
+		$this->setExpectedException( ParameterAssertionException::class );
+		new MultiplexingEntityLookup( $lookups );
+	}
+
+	public function provideInvalidForeignLookups() {
+		return array(
```
JeroenDeDauw: Such PHP 5.3!

src/Lookup/UnknownForeignRepositoryException.php
```diff
@@ -0,0 +1,42 @@
...
+			0,
+			$previous
+		);
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getRepoName() {
+		return $this->repoNname;
```
filbertkm: repoNname?

src/Lookup/MultiplexingEntityLookup.php
manicki: Renamed to `DispatchingEntityLookup` for now. Could come up with a nice way to include the repository name in the name (`RepositoryNameBasedDispatchingEntityLookup` is not the nicest name) but any suggestions are welcome!

manicki:
I've done what @JeroenDeDauw and @jakobw suggested (except changing the array syntax, will do this separately for the whole repository)

tests/unit/Lookup/MultiplexingEntityLookupTest.php
manicki: True but this seems to be syntax used in this repository. I am always confused what Wikibase repo uses what array syntax. I'll made a separate PR for changing this

src/Lookup/UnknownForeignRepositoryException.php
manicki: Forgot about this one. Thanks for spotting. Fixed now

JeroenDeDauw:
Should this still be DNM?

manicki:
maybe not but it still requires DM 6.2 release, so in this sense it is still kinda DNM. But removing the "label" now.

JeroenDeDauw:
Ah right. You don't really need that if the tests are not passing :)

src/Lookup/DispatchingEntityLookup.php
```diff
@@ -0,0 +1,92 @@
...
+	/**
+	 * @param EntityId $entityId
+	 *
+	 * @return EntityLookup
+	 * @throws UnknownForeignRepositoryException
+	 */
+	private function getLookupForEntityId( EntityId $entityId ) {
+		$repo = $entityId->getRepositoryName();
+		if ( !isset( $this->lookups[$repo] ) ) {
+			throw new UnknownForeignRepositoryException( $entityId );
```
jakobw: Just like in the other PR this should probably also be `$entityId->getRepositoryName()` instead

manicki: oops, right. Thanks!

manicki:
Code adding the new exception have been separated to https://github.com/wmde/WikibaseDataModelServices/pull/154 so other PRs could also use it!

manicki:
I've just realized that what I was trying to do here was wrong. Simple passing of things like `foo:Q123` to the lookup related to the repository locally known as `foo` does not seem right. That repo's database etc does not know a thing that some other repo calls it a `foo`.
What I believe should happen is that the dispatching lookup picks a right repo-specific lookup (that was done already in this patch), and then passes an entity id without a repo prefix to the right lookup.
I've added https://github.com/wmde/WikibaseDataModelServices/pull/150/commits/6669deb1433fa742cbc42980a19c2f504bf63d2d to do this.

Some "alternative" would be not to do this but rather adjust lookup code to not rely on the full id serialization but the form of it not including repo prefix. That seems logically incorrect for me, and also it would mean change some billion and a half places in Wikibase-related code to use `EntityId::getLocalPart` instead of `EntityId::getSerialization` so I decided not to go this path

jakobw:
> Some "alternative" would be not to do this but rather adjust lookup code to not rely on the full id serialization but the form of it not including repo prefix. That seems logically incorrect for me, and also it would mean change some billion and a half places in Wikibase-related code to use EntityId::getLocalPart instead of EntityId::getSerialization so I decided not to go this path

Could you expand on why it would be logically incorrect to keep the repo prefix? To me it seems sensible to pass foreign EntityIds to services that operate on foreign repositories. It sounds like a dangerous thing to convert them to "local" early on. This would make it impossible for the Lookup-service to know whether the EntityId it received really belongs to its repo. Not sure if that's really a concern though.

manicki:
> Could you expand on why it would be logically incorrect to keep the repo prefix? To me it seems sensible to pass foreign EntityIds to services that operate on foreign repositories. It sounds like a dangerous thing to convert them to "local" early on. 

For me `foo:Q123` reads like "an entity with id Q123 coming from the foreign repository I call foo". Following that, what I want to look up from the that repository (repo foo) is an entity identified as Q123. "foo" prefix only tells me I should be asking that repo, not my local repo.

Those repo prefixes only have a meaning inside the repo defining it. I believe it is not right to rely on them and refer to them outside of the scope of the "local" repo in any way.

> This would make it impossible for the Lookup-service to know whether the EntityId it received really belongs to its repo. Not sure if that's really a concern though.

I get this concern. But what if repo I call "foo" was using entities from some other repo, and by accident it would call the other repo also "foo". Then when my repo asks my "foo" for "foo:Q123" how the "my foo" repo knows is the question about its own Q123 or maybe the Q123 it got from the other foreign repo?

This whole process should be also taking the "prefix chaining" and "prefix mapping" into account. If repo I call "foo" sends me an entity "bar:Q321", and I have a mapping of "foo:bar" to "wd", I cannot ask repo foo for "wd:Q321" as it would not know if this is the request about its local Q321 or maybe the thing it knows as "bar:Q321" (to make the example even more horrible: let's assume repo foo used some entities from some other foreign repo and used "wd" for the prefix...). In such cases I think my repo should rather resolve "wd:Q321" first, and then ask repo "foo" for "bar:Q321".

jakobw:
Ah, I think we were misunderstanding each other. Of course we should not ask "foo" what `foo:Q123` is. The prefix definitely has to be removed at some point. My question was whether it should be the job of the dispatching service or the service that the dispatcher picked. To my mind it makes more sense to make it the responsibility of the service that actually interacts with "foo" and not the dispatcher but there might be good reasons to do it the other way.

manicki:
Oh, that's what you meant. Sorry I didn't get it.
You're making a good point actually. It indeed seems that there should be a decision that either all services should be able to handle the entity ids, no matter if they're foreign or not, or we assume that only _dispatching_ services deal with the foreign ids and always pass the "unprefixed" id to the right service.
At first glance I agree that it seems (to me) that the former would be nicer. On the other hand, that would most probably involve quite some change to many services. Not that should be seen as an argument against choosing that way.
I'd be happy to remove this change from this patch. It would be still worth looking deeper also into services we haven't touched yet. Some other opinions would be also appreciated.

It looks like would would also need a service that would handle the dropping of a prefixing and (especially) handle the prefix unmapping, and depending on the decision, use it in _all_ services or in the dispatching services.

brightbyte:
Regarding the removal of the prefix: it's indeed not a trivial question. It boils down to whether the repo-specific service instances should "think like" the foreign repo, or like the local repo. Let's call this the "prefix resolution context".

Pragmatically, I propose the following:
- The dispatching services does not strip the prefix, it passes the EntityId to the underlying service as-is. This way, we can use the same API service instance for all repos we want to access via API, if the implementation looks up the API URL based on the prefix. This also allows us to stack multiple dispatching services (though I'm not sure we need this).
- service concrete implementations may be bound to a specific repo (they know a database name or API URL). Such implementations should check the prefix, and fail if expectations are not met.
- There could be a wrapper implementation of the service that does prefix stripping and/or un-mapping. If the service methods returns any EntityIds, prefix mapping should be applied, to maintain symmetry.
- A DB based implementation may only support local IDs, since it uses the serialized IDs as field values. To use that implementation with the DB of a foreign repo, it would need to be wrapped, to remove prefixed from input, and add prefixes to the output.
- when stripping prefixes, only the first part can be stripped. When processing "foo:bar:Q5", "foo" should be asked for "bar:Q5".

The mapping/unmapping wrapper gives us full flexibility. But it's quite a bit of boiler-plate code for each service interface. We should only do this if we have an implementation that is not "prefix-aware". The documentation of the interface should mention that implementations must clearly document whether they use the local context for prefixes, or a foreign one, or whether they support no prefixes.

src/Lookup/DispatchingEntityLookup.php
```diff
@@ -0,0 +1,105 @@
...
+	/**
+	 * @param EntityLookup[] $lookups associative array with repository names (strings) as keys
+	 *                                and EntityLookup objects as values. Empty-string key
+	 *                                defines lookup for the local repository.
+	 *
+	 * @throws ParameterAssertionException
+	 */
+	public function __construct( array $lookups, EntityIdParser $parser ) {
+		Assert::parameter(
+			!empty( $lookups ) && array_key_exists( '', $lookups ),
```
brightbyte: I see no reason to require this here. The dispatcher shouldn't know or care that the empty prefix is somehow special.

To protect against bad configuration, the code that calls the constructor can do this check. That's bootsrap code that knows about configuration and the context in which the instance will be used. So it can know whether a handler for local IDs is needed.

src/Lookup/DispatchingEntityLookup.php
```diff
@@ -0,0 +1,105 @@
...
+			'$lookups',
+			'must must not be empty and must contain an empty-string key'
+		);
+		Assert::parameterElementType( EntityLookup::class, $lookups, '$lookups' );
+		Assert::parameterElementType( 'string', array_keys( $lookups ), 'array_keys( $lookups )' );
+		foreach ( array_keys( $lookups ) as $repositoryName ) {
```
brightbyte: Maybe we want a convenience function for this in some helper somewhere.

src/Lookup/DispatchingEntityLookup.php
```diff
@@ -0,0 +1,105 @@
...
+	 *
+	 * @return null|EntityDocument
+	 * @throws EntityLookupException
+	 * @throws UnknownForeignRepositoryException
+	 */
+	public function getEntity( EntityId $entityId ) {
+		$lookup = $this->getLookupForEntityId( $entityId );
+		if ( $entityId->isForeign() ) {
+			$entityId = $this->parser->parse( $entityId->getLocalPart() );
+		}
```
brightbyte: The dispatcher should not strip prefixes, see my post to the conversation thread. 
Since it's not obvious what the caller can expect, the documentation of this class should be explicit about whether prefixes are stripped or not.

src/Lookup/DispatchingEntityLookup.php
```diff
@@ -0,0 +1,105 @@
...
+
+use Wikibase\DataModel\Entity\EntityDocument;
+use Wikibase\DataModel\Entity\EntityId;
+use Wikibase\DataModel\Entity\EntityIdParser;
+use Wikimedia\Assert\Assert;
+use Wikimedia\Assert\ParameterAssertionException;
+
+/**
+ * Delegates lookup to the repository-specific EntityLookup
+ * based on the name of the repository EntityId is coming from.
```
brightbyte: ...based on the name of the repository an EntityId belongs to.

src/Lookup/DispatchingEntityLookup.php
```diff
@@ -0,0 +1,105 @@
...
+	 * @return null|EntityDocument
+	 * @throws EntityLookupException
+	 * @throws UnknownForeignRepositoryException
+	 */
+	public function getEntity( EntityId $entityId ) {
+		$lookup = $this->getLookupForEntityId( $entityId );
+		if ( $entityId->isForeign() ) {
+			$entityId = $this->parser->parse( $entityId->getLocalPart() );
+		}
+		return $lookup->getEntity( $entityId );
```
brightbyte, line 60: If we do prefix stripping above, the ID _inside_ the Entity would need to be changed to include the prefix now. It's important to maintain symmetry in this regard, so the caller only has to deal with the local "prefix resolution context".

brightbyte (changes requested):
Dispatcher should not strip prefixes

src/Lookup/DispatchingEntityLookup.php
manicki: done

src/Lookup/DispatchingEntityLookup.php
manicki: Removed the requirement for the empty-string key

src/Lookup/DispatchingEntityLookup.php
manicki: I've submitted something for this as https://github.com/wmde/WikibaseDataModelServices/pull/155.
To avoid creating yet more mess with all commits in this PR, I'd rather wait for that PR to at least be in shape ready to be merged and then do changes here (and in other places). 

Please do not "approve the change" (do not remove -1 in github's speak) before it is done so we don't forget about this.

src/Lookup/DispatchingEntityLookup.php
manicki: Removed the code stripping prefixes

brightbyte:
The code looks good now. As requested, I'm holding back on approving/merging until #155 is merged, and can be used here.

Also, please be aware that the "inner" EntityLookup must use the correct ID mapping when unserializing entities from another repository. This probably needs to be implemented in WikiPageEntityRedirectLookup / EntityContentDataCodec.

addshore:
As with my 2p on https://github.com/wmde/WikibaseDataModelServices/pull/149 I have the same 2p here, that both classes and methods should have @since tags. 4 years down the line this makes everyone a bunch easier! and all for only the cost of a few lines!

manicki:
Added since tags to public methods as well, and switched the new class to use validation function from DM 6.3.
As this PR has been getting horrendous I have squashed 20+ commits into one.
To verify I only changed two things mentioned above, one could run `git diff -r 5b8bb0a78c1656d24c4fd7bb1ae406ae0b265843 a8d7457a5c32fdd21a8a1f7c6cce5be69d130bfe` locally (watch out the rebase-related noise)

brightbyte (approved):

PR #151, opened 2016-10-12
Add DispatchingTermLookup

jakobw:
- adds a service that picks a `TermLookup` for the corresponding repository for a given `EntityId`
- implemented on top of #150 in order to reuse `UnknownForeignRepositoryException`

Task on Phabricator: https://phabricator.wikimedia.org/T149583

### DIFF
--- a/src/Lookup/DispatchingTermLookup.php
+++ b/src/Lookup/DispatchingTermLookup.php
@@ -113,7 +113,7 @@ class DispatchingTermLookup implements TermLookup {
 	private function getLookupForEntityId( EntityId $entityId ) {
 		$repo = $entityId->getRepositoryName();
 		if ( !isset( $this->lookups[$repo] ) ) {
-			throw new UnknownForeignRepositoryException( $entityId->getRepositoryName() );
+			throw new UnknownForeignRepositoryException( $repo );
 		}
 		return $this->lookups[$repo];
 	}
--- a/src/Lookup/DispatchingTermLookup.php
+++ b/src/Lookup/DispatchingTermLookup.php
@@ -70,7 +70,7 @@ class DispatchingTermLookup implements TermLookup {
 	 * @throws TermLookupException
 	 * @throws UnknownForeignRepositoryException
 	 *
-	 * @return null|string
+	 * @return string[]
 	 */
 	public function getLabels( EntityId $entityId, array $languageCodes ) {
 		return $this->getLookupForEntityId( $entityId )->getLabels( $entityId, $languageCodes );
@@ -91,7 +91,6 @@ class DispatchingTermLookup implements TermLookup {
 		return $this->getLookupForEntityId( $entityId )->getDescription( $entityId, $languageCode );
 	}
 
-
 	/**
 	 * @see TermLookup::getDescriptions
 	 *
@@ -101,7 +100,7 @@ class DispatchingTermLookup implements TermLookup {
 	 * @throws TermLookupException
 	 * @throws UnknownForeignRepositoryException
 	 *
-	 * @return null|string
+	 * @return string[]
 	 */
 	public function getDescriptions( EntityId $entityId, array $languageCodes ) {
 		return $this->getLookupForEntityId( $entityId )->getDescriptions( $entityId, $languageCodes );
@@ -118,4 +117,5 @@ class DispatchingTermLookup implements TermLookup {
 		}
 		return $this->lookups[$repo];
 	}
+
 }
--- a/src/Lookup/DispatchingTermLookup.php
+++ b/src/Lookup/DispatchingTermLookup.php
@@ -113,7 +113,7 @@ class DispatchingTermLookup implements TermLookup {
 	private function getLookupForEntityId( EntityId $entityId ) {
 		$repo = $entityId->getRepositoryName();
 		if ( !isset( $this->lookups[$repo] ) ) {
-			throw new UnknownForeignRepositoryException( $entityId, 'Unknown repository: ' . $repo );
+			throw new UnknownForeignRepositoryException( $entityId );
 		}
 		return $this->lookups[$repo];
 	}
--- a/src/Lookup/DispatchingTermLookup.php
+++ b/src/Lookup/DispatchingTermLookup.php
@@ -23,8 +23,8 @@ class DispatchingTermLookup implements TermLookup {
 
 	/**
 	 * @param TermLookup[] $lookups associative array with repository names (strings) as keys
-	 *                                and TermLookup objects as values. Empty-string key
-	 *                                defines lookup for the local repository.
+	 *                              and TermLookup objects as values. Empty-string key
+	 *                              defines lookup for the local repository.
 	 *
 	 * @throws ParameterAssertionException
 	 */
--- a/src/Lookup/DispatchingTermLookup.php
+++ b/src/Lookup/DispatchingTermLookup.php
@@ -29,11 +29,7 @@ class DispatchingTermLookup implements TermLookup {
 	 * @throws ParameterAssertionException
 	 */
 	public function __construct( array $lookups ) {
-		Assert::parameter(
-			!empty( $lookups ) && array_key_exists( '', $lookups ),
-			'$lookups',
-			'must must not be empty and must contain an empty-string key'
-		);
+		Assert::parameter( !empty( $lookups ), '$lookups', 'must must not be empty' );
 		Assert::parameterElementType( TermLookup::class, $lookups, '$lookups' );
 		Assert::parameterElementType( 'string', array_keys( $lookups ), 'array_keys( $lookups )' );
 		foreach ( array_keys( $lookups ) as $repositoryName ) {
--- a/src/Lookup/DispatchingTermLookup.php
+++ b/src/Lookup/DispatchingTermLookup.php
@@ -113,7 +113,7 @@ class DispatchingTermLookup implements TermLookup {
 	private function getLookupForEntityId( EntityId $entityId ) {
 		$repo = $entityId->getRepositoryName();
 		if ( !isset( $this->lookups[$repo] ) ) {
-			throw new UnknownForeignRepositoryException( $entityId );
+			throw new UnknownForeignRepositoryException( $entityId->getRepositoryName() );
 		}
 		return $this->lookups[$repo];
 	}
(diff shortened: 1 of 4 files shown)


src/Lookup/DispatchingTermLookup.php
```diff
@@ -0,0 +1,121 @@
...
+class DispatchingTermLookup implements TermLookup {
+
+	/**
+	 * @var TermLookup[]
+	 */
+	private $lookups;
+
+	/**
+	 * @param TermLookup[] $lookups associative array with repository names (strings) as keys
+	 *                                and TermLookup objects as values. Empty-string key
```
manicki: indentation is a bit off here and line below

src/Lookup/DispatchingTermLookup.php
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	/**
+	 * @param EntityId $entityId
+	 * @return TermLookup
+	 */
+	private function getLookupForEntityId( EntityId $entityId ) {
+		$repo = $entityId->getRepositoryName();
+		if ( !isset( $this->lookups[$repo] ) ) {
+			throw new UnknownForeignRepositoryException( $entityId, 'Unknown repository: ' . $repo );
```
manicki: Just realized that the second parameter passed to the constructor could be possibly skipped in all dispatching service patches as it is not really relevant, ie. ''Unknown repository name: foo" vs "'Unknown repository: foo".
I'll remove it my patches. If shorter message is preferred, it should be changed in the exception. I have no preference here.

tests/unit/Lookup/DispatchingTermLookupTest.php
```diff
@@ -0,0 +1,204 @@
...
+	 */
+	public function testGivenEntityFromUnknownRepo_throwsException( $method, $language ) {
+		$item = new ItemId( 'fooo:Q123' );
+		$lookup = new DispatchingTermLookup( [
+			'' => $this->getMock( TermLookup::class ),
+			'foo' => $this->getMock( TermLookup::class ),
+		] );
+
+		$this->setExpectedException( UnknownForeignRepositoryException::class );
+		$lookup->$method( $item, $language );
```
manicki, line 57: Let's see if we will have @JeroenDeDauw ranting on this line :)

manicki (changes requested):
Looks good at glance, generally follows the way other dispatching services have been proposed so far.

Found two minor things, one being my mistake actually.

I see you have separate tests for "entity from local repo" and "entity from foreign repo" cases. I am a bit torn on this (I originally also made such division in `DispatchingEntityLookup`'s tests). On one hand local repo case is a bit special and different from the foreign entity case. On the other hand, technically local repo and any foreign repo are pretty much the same thing for the dispatcher, so I started to think that maybe it would be enough to just check in tests that some things work for the known repo (be it local or foreign, it does not matter), and that for unknown repo exceptions are thrown etc.
I'd really appreciate other's opinions on this. We're getting more and more of those dispatchers, so it would be nice to have tests consistent at the early stage, I think.

According to docs all public methods of `TermLookup` throw `TermLookupException`. Those exception should not be caught be the dispatching wrapper. From looking at the code I can tell it does not happen but some small unit tests proving this would be nice too.

src/Lookup/DispatchingTermLookup.php
jakobw: Done!

src/Lookup/DispatchingTermLookup.php
jakobw: Removed!

tests/unit/Lookup/DispatchingTermLookupTest.php
jakobw: :)
I was too lazy to write the same thing 4x but I'm fine with changing it.

jakobw:
Thanks @manicki for the review!

I agree with what you said about testing the two cases for each method and would also like to hear a third opinion on it. I have no strong preference.

Regarding `TermLookupException` I felt like it would be enough to have the `@throws`-tag and have people trust that it won't secretly be caught by the dispatching service. I'd also be fine with adding a test though.

manicki (approved):

manicki:
Thanks for quick fixes!

src/Lookup/DispatchingEntityLookup.php
```diff
@@ -0,0 +1,92 @@
+<?php
```
mariushoch: I don't think this should be part of this PR, should it?

src/Lookup/DispatchingTermLookup.php
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	/**
+	 * @param EntityId $entityId
+	 * @return TermLookup
+	 */
+	private function getLookupForEntityId( EntityId $entityId ) {
+		$repo = $entityId->getRepositoryName();
+		if ( !isset( $this->lookups[$repo] ) ) {
+			throw new UnknownForeignRepositoryException( $entityId );
```
mariushoch: First parameter to the constructor of `UnknownForeignRepositoryException` is `string $repoName` not an entity id.

mariushoch:
Not sure what happened to this PR, but it seems there are to many commits and way to many changes in here.

src/Lookup/DispatchingTermLookup.php
jakobw: Indeed, thanks!

jakobw:
@mariushoch the PR looks like that because it's depending on @manicki's PR #150 and includes all of its commits.

jakobw:
Rebased against #154 to avoid all the commits from #150 flying around.

src/Lookup/DispatchingTermLookup.php
```diff
@@ -0,0 +1,121 @@
...
+	 *                              and TermLookup objects as values. Empty-string key
+	 *                              defines lookup for the local repository.
+	 *
+	 * @throws ParameterAssertionException
+	 */
+	public function __construct( array $lookups ) {
+		Assert::parameter(
+			!empty( $lookups ) && array_key_exists( '', $lookups ),
+			'$lookups',
+			'must must not be empty and must contain an empty-string key'
```
brightbyte: Why is that required? Do we rely on that anywhere?

tests/unit/Lookup/DispatchingTermLookupTest.php
brightbyte: I think for the negative test, this is ok.

tests/unit/Lookup/DispatchingTermLookupTest.php
```diff
@@ -0,0 +1,204 @@
...
+	public function publicMethodsProvider() {
+		return [
+			[ 'getLabel', 'en' ],
+			[ 'getLabels', [ 'en', 'fr' ] ],
+			[ 'getDescription', 'en' ],
+			[ 'getDescription', [ 'en', 'de' ] ],
+		];
+	}
+
+	public function testGetLabelFromLocalRepo() {
```
brightbyte: The dispatcher should not really know or care about local and foreign. It uses whatever prefix is in the ID - if the prefix is the empty string, it should just work like any other. That should be tested, but it doesn't need a separate function.

brightbyte (changes requested):
Looks pretty good, some small requests

src/Lookup/DispatchingTermLookup.php
mariushoch: Not really, but that would require changes to the error handling so that it can handle that special case in a meaningful way.
In general, I could see that you want to treat all lookups equally "remote"… but I don't think that matches up with what was done in DM, does it?

brightbyte (approved):

brightbyte:
I'm confused about the Travis failures. It seems that it is using the old DataModel version for some test runs, but not for others... is there a way to re-trigger the test runs?

brightbyte:
Except for the Travis failures, I think this is good to merge now.

jakobw:
@brightbyte travis config is fixed now.

src/Lookup/DispatchingTermLookup.php
```diff
@@ -0,0 +1,117 @@
...
+	 *                              and TermLookup objects as values. Empty-string key
+	 *                              defines lookup for the local repository.
+	 *
+	 * @throws ParameterAssertionException
+	 */
+	public function __construct( array $lookups ) {
+		Assert::parameter( !empty( $lookups ), '$lookups', 'must must not be empty' );
+		Assert::parameterElementType( TermLookup::class, $lookups, '$lookups' );
+		Assert::parameterElementType( 'string', array_keys( $lookups ), 'array_keys( $lookups )' );
+		foreach ( array_keys( $lookups ) as $repositoryName ) {
```
brightbyte, line 35: should probably wait for #155, so we can get rid of this.

brightbyte (changes requested):

brightbyte:
@jakobw do we have a ticket for this somewhere?

jakobw:
@brightbyte Not yet, I'll create one.

tests/unit/Lookup/DispatchingTermLookupTest.php
```diff
@@ -0,0 +1,177 @@
...
+	public function publicMethodsProvider() {
+		return [
+			[ 'getLabel', 'en' ],
+			[ 'getLabels', [ 'en', 'fr' ] ],
+			[ 'getDescription', 'en' ],
+			[ 'getDescription', [ 'en', 'de' ] ],
+		];
+	}
+
+	public function testGetLabel() {
```
JeroenDeDauw, line 69: Not a very descriptive name

manicki:
As said above, the code is looking pretty good. I am just having some problems now to see this service being used in `WikibaseClient`: https://phabricator.wikimedia.org/diffusion/EWBA/browse/master/client/includes/WikibaseClient.php
It seems to be that client mostly uses the `BufferingTermLookup`: https://phabricator.wikimedia.org/diffusion/EWBA/browse/master/lib/includes/Store/BufferingTermLookup.php
So I am now thinking do we want a separate buffering term lookup for each repository, or maybe rather a single buffer for multiple term lookups/term indexes? If the former is the case, we would probably rather need a dispatching buffering lookup service then?

I am not sure if this PR is a right place to discuss such issues, if not let's have it somewhere else.

brightbyte:
As I said in https://phabricator.wikimedia.org/T148141#2760606, we indeed do need a DispatchingTermBuffer (which is also a dispatching TermLookup), but probably no "plain" DispatchingTermLookup. Sorry about that...

jakobw:
Closed for the reasons mentioned in the comment above

PR #153, opened 2016-10-13
Add EntityIdPager from Wikibase\Repo\Store

mariushoch:
Simple interface for services that allow paging through EntityIds.

### DIFF
--- a/src/EntityId/EntityIdPager.php
+++ b/src/EntityId/EntityIdPager.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Wikibase\DataModel\Services\EntityId;
+
+use Wikibase\DataModel\Entity\EntityId;
+
+/**
+ * A cursor for paging through EntityIds.
+ *
+ * @since 3.7
+ *
+ * @license GPL-2.0+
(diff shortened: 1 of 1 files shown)


src/EntityId/EntityIdPager.php
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * A cursor for paging through EntityIds.
+ *
+ * @since 3.7
+ *
+ * @license GPL-2.0+
+ * @author Daniel Kinzler
+ */
+interface EntityIdPager {
```
JeroenDeDauw, line 15: Might be worth changing so it implements the Fetcher interface from https://github.com/JeroenDeDauw/BatchingIterator, which then allows for having an Iterator and hiding that a fetcher is used

JeroenDeDauw (approved):
+1, thanks for moving

src/EntityId/EntityIdPager.php
mariushoch: I'm open to that, but let's see what others say. Also means a new dependency.

PR #154, opened 2016-10-14
Add UnknownForeignRepositoryException

manicki:
This is taken out of https://github.com/wmde/WikibaseDataModelServices/pull/150 so other PRs needing the exception don't have to include all other things from that PR (e.g. https://github.com/wmde/WikibaseDataModelServices/pull/151)

### DIFF
--- a/src/Lookup/UnknownForeignRepositoryException.php
+++ b/src/Lookup/UnknownForeignRepositoryException.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Lookup;
+
+use Exception;
+use InvalidArgumentException;
+
+/**
+ * @since 3.7
+ *
+ * @license GPL-2.0+
+ */
+class UnknownForeignRepositoryException extends InvalidArgumentException {
+
+	/**
+	 * @var string
+	 */
+	private $repositoryName;
+
+	/**
+	 * @param string $repositoryName
+	 * @param string|null $message
+	 * @param Exception|null $previous
+	 */
+	public function __construct( $repositoryName, $message = null, Exception $previous = null ) {
+		$this->repositoryName = $repositoryName;
+
+		parent::__construct(
+			$message ?: 'Unknown repository name: ' . $repositoryName,
+			0,
+			$previous
+		);
+	}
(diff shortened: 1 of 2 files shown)


jakobw (approved):
Looks good to me!

src/Lookup/UnknownForeignRepositoryException.php
```diff
@@ -0,0 +1,42 @@
...
+			$message ?: 'Unknown repository name: ' . $repoName,
+			0,
+			$previous
+		);
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getRepoName() {
```
thiemowmde: Didn't we decided to call this `getRepositoryName` in other classes? This should be consistent, please.

tests/unit/Lookup/UnknownForeignRepositoryExceptionTest.php
```diff
@@ -0,0 +1,34 @@
...
+		$exception = new UnknownForeignRepositoryException( 'foo' );
+
+		$this->assertSame( 'foo', $exception->getRepoName() );
+		$this->assertSame( 'Unknown repository name: foo', $exception->getMessage() );
+		$this->assertSame( 0, $exception->getCode() );
+		$this->assertNull( $exception->getPrevious() );
+	}
+
+	public function testConstructWithAllArguments() {
+		$previous = new Exception( 'Dummy exception' );
```
thiemowmde: The dummy string is not necessary, it does not add value to the test. It's even a bit confusing to call this a "dummy", because it is none. Could be removed.

thiemowmde (changes requested):

src/Lookup/UnknownForeignRepositoryException.php
manicki: True. Sorry for being sloppy

tests/unit/Lookup/UnknownForeignRepositoryExceptionTest.php
manicki: Removed

thiemowmde (approved):

PR #155, opened 2016-10-31
Add functions to assert a value is an allowable repository name, or an array indexed by repository names

manicki:
Such checks have been done in https://github.com/wmde/WikibaseDataModelServices/pull/150, https://github.com/wmde/WikibaseDataModelServices/pull/151 or https://gerrit.wikimedia.org/r/#/c/314682/, as well as in https://github.com/wmde/WikibaseDataModelServices/pull/146 and https://github.com/wmde/WikibaseDataModelServices/pull/149. Let's have a dedicated function/functions for this.

I am not sure about the name/location of the new class, and function names are horrible. Suggestions would be great!

Also I don't know if we want to have dedicated exception(s) for those rather special cases.

### DIFF
--- a/src/Assert/Assert.php
+++ b/src/Assert/Assert.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Assert;
+
+use Wikimedia\Assert\Assert as WikimediaAssert;
+use Wikimedia\Assert\ParameterAssertionException;
+
+/**
+ * Provides functions to assure values meet certain preconditions relevant
+ * in Wikibase.
+ *
+ * @see Wikimedia\Assert\Assert
+ *
+ * @since 3.7
+ *
+ * @license GPL-2.0+
+ */
+class Assert {
+
+	/**
+	 * @param string $value
+	 * @param string $name
+	 *
+	 * @throws ParameterAssertionException if $value is not an allowable repository name.
+	 */
+	public static function parameterIsAllowableRepositoryName( $value, $name ) {
+		if ( !self::isAllowableRepositoryName( $value ) ) {
+			throw new ParameterAssertionException( $name, 'must be a string not including colons' );
+		}
+	}
+
+	/**
+	 * @param array $value
+	 * @param string $name
+	 *
+	 * @throws ParameterAssertionException If an element of $value is not
+	 *         an allowable repository name.
+	 */
+	public static function parameterKeyIsAllowableRepositoryName( $value, $name ) {
+		WikimediaAssert::parameterType( 'array', $value, $name );
+		// TODO: change to Assert::parameterKeyType when the new version of the library is released
+		WikimediaAssert::parameterElementType( 'string', array_keys( $value ), "array_keys( $name )" );
+
+		foreach ( array_keys( $value ) as $key ) {
+			if ( !self::isAllowableRepositoryName( $key ) ) {
+				throw new ParameterAssertionException(
+					"array_keys( $name )",
+					'must not contain strings including colons'
+				);
+			}
+		}
+	}
+
+	/**
+	 * @param string $value
+	 * @return bool
+	 */
+	private static function isAllowableRepositoryName( $value ) {
+		return is_string( $value ) && strpos( $value, ':' ) === false;
+	}
+
+}
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
 		"php": ">=5.5.0",
 		"wikibase/data-model": "~6.0|~5.0|~4.2",
 		"data-values/data-values": "~0.1|~1.0",
-		"diff/diff": "~2.0|~1.0"
+		"diff/diff": "~2.0|~1.0",
+		"wikimedia/assert": "~0.2.2"
 	},
 	"require-dev": {
 		"squizlabs/php_codesniffer": "~2.3",
--- a/tests/unit/Assert/AssertTest.php
+++ b/tests/unit/Assert/AssertTest.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Tests\Assert;
+
+use Wikibase\DataModel\Services\Assert\Assert;
+use Wikimedia\Assert\ParameterAssertionException;
+
+/**
+ * @covers Wikibase\DataModel\Services\Assert\Assert
+ *
+ * @license GPL-2.0+
+ */
+class AssertTest extends \PHPUnit_Framework_TestCase {
+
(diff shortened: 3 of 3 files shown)


src/Assert/Assert.php
```diff
@@ -0,0 +1,62 @@
...
+ */
+class Assert {
+
+	/**
+	 * @param string $value
+	 * @param string $name
+	 *
+	 * @throws ParameterAssertionException if $value is not an allowable repository name.
+	 */
+	public static function parameterIsAllowableRepositoryName( $value, $name ) {
```
jakobw, line 26: `parameterIsValidRepositoryName` sounds slightly better to me. Just `isValidRepositoryName` sounds even better but I guess that's not in line with the current names of the methods in `Assert`.

src/Assert/Assert.php
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	/**
+	 * @param array $value
+	 * @param string $name
+	 *
+	 * @throws ParameterAssertionException If an element of $value is not
+	 *         an allowable repository name.
+	 */
+	public static function parameterKeyIsAllowableRepositoryName( $value, $name ) {
```
jakobw, line 39: Again, I think "valid" sounds better.

brightbyte:
I'm fine with merging this as it is. Leaving it until tomorrow to give others a chance to comment.

src/Assert/Assert.php
```diff
@@ -0,0 +1,62 @@
...
+ * Provides functions to assure values meet certain preconditions relevant
+ * in Wikibase.
+ *
+ * @see Wikimedia\Assert\Assert
+ *
+ * @since 3.7
+ *
+ * @license GPL-2.0+
+ */
+class Assert {
```
jakobw, line 18: I'm not quite sure how I feel about giving this class such a generic name. I can see that its role is to contain methods that would be too specific for `Wikimedia\Assert` but it seems very procedural style to have a class like this with methods for very specific services that get used all over the codebase.

jakobw (commented):
Thanks for tackling this!
Added some comments, but I'd also be fine with it as is.

src/Assert/Assert.php
brightbyte: isFoo should return a bool. For things that throw exceptions, the verb should be assert or ensure.

src/Assert/Assert.php
brightbyte: Agreed.

src/Assert/Assert.php
brightbyte: Yea... what other methods may we want to add here? If we had something to validate, say, site IDs, should it be in the same class? That kind of feels wrong...

Actually, the syntax is defined in the DataModel module, not in DataModelServices... so perhaps the assertions should be in DataModel too?

JeroenDeDauw: You could just create some global functions, no need to pretend this code is OO`</grump>`

src/Assert/Assert.php
```diff
@@ -0,0 +1,62 @@
...
+	 */
+	public static function parameterIsAllowableRepositoryName( $value, $name ) {
+		if ( !self::isAllowableRepositoryName( $value ) ) {
+			throw new ParameterAssertionException( $name, 'must be a string not including colons' );
+		}
+	}
+
+	/**
+	 * @param array $value
+	 * @param string $name
```
JeroenDeDauw, line 34: $value is a weird name given this is an array

I don't understand what `$name` is, even after looking at the implementation

JeroenDeDauw (changes requested):
Even more static code...

How about creating a RepositoryName value object?

src/Assert/Assert.php
manicki: Yeah, names. So I initially also thought about "valid" but it also felt like "valid repository name" could be understood as the "repository name of a foreign repository that has been configured" etc. while this is just about a name format.
But yeah, maybe it is over-thinking things. Will change to "assertParameterIsValidRepositoryName".

src/Assert/Assert.php
manicki: Well, I kinda re-used the interface of wikimedia/Assert: https://github.com/wikimedia/Assert/blob/master/src/Assert.php
$name is a name of the parameter. It is passed over to the exceptions thrown.

src/Assert/Assert.php
manicki: So I first called the class "DispatchingServicesAssert" as it would be used for this "family" of services. But then realized it could also be used elsewhere. So I changed the name to "RepositoryNameAssert" for a second but this felt a bit silly so I shortened it.
I agree that functions validating some other things should not go in here. So the current class name is indeed too generic. I'll change it to something more "specialized".
And actually putting it to DataModule might make sense. I first thought let's have it in DMServices as things from this module are going to use it but as said, the syntax is in DM so validation is more naturally belonging here.

Global functions - It makes sense. I just made this fake-OO thing as the code using validation functions defined here also uses some other functions from wikimedia/Assert that use this OO-like interface. So I didn't want to have something different and confuse users. But true, this is a weak argument.

manicki:
I am not sure about the value object approach. As this is only intended to validate parameters in constructors of some services I am not sure wouldn't this be over-designing things.

manicki:
Thanks for comments @jakob-WMDE @brightbyte @JeroenDeDauw . As suggested those functions fit better in DataModel module, so I am closing this pull request and will re-submit it in the other component, including changes suggested already here in the review.

src/Assert/Assert.php
brightbyte: @JeroenDeDauw Assert is purely static, since it's basically a language extension - an alternative to the native (broken) assert() function. At the time, there was some reluctance to adopt namespaced top level functions  for this purpose. To me, that's a cosmetic difference.

The functionality proposed here is much less related to language features; it is in fact bound to the contract of EntityId::splitSerialization. It might be best to also have the assertion in EntityId. It could also be in a trait.

The syntax definition is currently bound to the EntityId class in a static way - it is not configurable or injectable. For that reason, the assertion does not need to be injectable or configurable either. I'm a bit reluctant to put more static stuff into EntityId, but logically, there is no reason not to.

src/Assert/Assert.php
brightbyte: "valid" is ok here. If you want to follow the example of XML, you could used "well formed" as the predicate.

brightbyte:
@JeroenDeDauw wrote:

> How about creating a RepositoryName value object?

Hell no. We have enough performance problems with loads-of-little-objects during de-derialization. I don't really want to double that.

PR #157, opened 2016-11-07
Add StatementListChanger

thiemowmde:
* ~~This wraps the last remaining usage of the ByPropertyIdArray class. This is meant as a migration path. First, all relevant tests from ByPropertyIdArrayTest should be carried over. Then the implementation can be changed to not rely on ByPropertyIdArray any more. ByPropertyIdArray can be removed then.~~
* I decided to **not** reimplement certain features of ByPropertyIdArray, especially it's magic behavior of reordering groups.
* This also adds a few service methods that are already in use in different places in the Wikibase.git code base, but with individual private implementations.

### DIFF
--- a/src/Statement/StatementListChanger.php
+++ b/src/Statement/StatementListChanger.php
@@ -1,0 +1,174 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Statement;
+
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Statement\Statement;
+use Wikibase\DataModel\Statement\StatementList;
+
+/**
+ * A collection of higher level utility functions to manipulate StatementList objects.
+ *
+ * @since 3.7
+ *
+ * @license GPL-2.0+
+ * @author Thiemo Mättig
+ */
+class StatementListChanger {
+
+	/**
+	 * @var StatementList
+	 */
+	private $statementList;
+
+	public function __construct( StatementList $statementList ) {
+		$this->statementList = $statementList;
+	}
+
+	/**
+	 * Makes sure all statements with the same property are next to each other (forming a group),
+	 * and reorders them if necessary. The position of the group in the list is determined by the
+	 * first statement with the same property.
+	 */
+	public function groupByProperty() {
+		$byId = [];
+
+		foreach ( $this->statementList->toArray() as $statement ) {
+			$id = $statement->getPropertyId()->getSerialization();
+			$byId[$id][] = $statement;
+		}
+
+		// FIXME: Use StatementList::clear, see https://github.com/wmde/WikibaseDataModel/pull/649
+		$this->clear();
+
+		foreach ( $byId as $statements ) {
+			foreach ( $statements as $statement ) {
+				$this->statementList->addStatement( $statement );
+			}
+		}
+	}
+
+	private function clear() {
+		foreach ( $this->statementList->toArray() as $statement ) {
+			$this->statementList->removeStatementsWithGuid( $statement->getGuid() );
+		}
+	}
+
+	/**
+	 * Adds a new statement to the list while respecting existing PropertyId groups.
+	 *
+	 * In contrast to the previously used ByPropertyIdArray implementation this method never moves
+	 * existing statements around. The provided index is only a hint. The new statement is either
+	 * added to an existing group of statements with the same PropertyId, or at a group border close
+	 * to the provided index.
+	 *
+	 * @param Statement $newStatement
+	 * @param int|null $index An absolute index in the list. If the index is not next to a statement
+	 *  with the same property, the closest possible position is used instead. Default is null,
+	 *  which adds the new statement after the last statement with the same property, or to the end.
+	 */
+	public function addToGroup( Statement $newStatement, $index = null ) {
+		$statements = $this->statementList->toArray();
+		$id = $newStatement->getPropertyId();
+
+		if ( $index === null ) {
+			$index = $this->getLastIndexWithinGroup( $statements, $id );
+		} else {
+			// Limit search range to avoid looping non-existing positions
+			$validIndex = min( max( 0, $index ), count( $statements ) );
+			$index = $this->getClosestIndexWithinGroup( $statements, $id, $validIndex );
+			if ( $index === null ) {
+				$index = $this->getClosestIndexAtGroupBorder( $statements, $validIndex );
+			}
+		}
+
+		$this->statementList->addStatement( $newStatement, $index );
+	}
+
+	/**
+	 * @param Statement[] $statements
+	 * @param PropertyId $id
+	 *
+	 * @return int|null
+	 */
+	private function getLastIndexWithinGroup( array $statements, PropertyId $id ) {
+		// Start searching from the end and stop at the first match
+		for ( $i = count( $statements ); $i > 0; $i-- ) {
+			if ( $statements[$i - 1]->getPropertyId()->equals( $id ) ) {
+				return $i;
+			}
+		}
+
+		return null;
+	}
+
+	/**
+	 * @param Statement[] $statements
+	 * @param PropertyId $id
+	 * @param int $index
+	 *
+	 * @return int|null
+	 */
+	private function getClosestIndexWithinGroup( array $statements, PropertyId $id, $index ) {
+		$longestDistance = max( $index, count( $statements ) - $index );
+
+		for ( $i = 0; $i <= $longestDistance; $i++ ) {
+			if ( $this->isWithinGroup( $statements, $id, $index - $i ) ) {
+				return $index - $i;
+			} elseif ( $i && $this->isWithinGroup( $statements, $id, $index + $i ) ) {
+				return $index + $i;
+			}
+		}
+
+		return null;
+	}
+
+	/**
+	 * @param Statement[] $statements
+	 * @param int $index
+	 *
+	 * @return int|null
+	 */
+	private function getClosestIndexAtGroupBorder( array $statements, $index ) {
+		$longestDistance = max( $index, count( $statements ) - $index );
+
+		for ( $i = 0; $i <= $longestDistance; $i++ ) {
+			if ( $this->isGroupBorder( $statements, $index - $i ) ) {
+				return $index - $i;
+			} elseif ( $i && $this->isGroupBorder( $statements, $index + $i ) ) {
+				return $index + $i;
+			}
+		}
+
+		return null;
+	}
+
+	/**
+	 * @param Statement[] $statements
+	 * @param PropertyId $id
+	 * @param int $index
+	 *
+	 * @return bool
+	 */
+	private function isWithinGroup( array $statements, PropertyId $id, $index ) {
+		$count = count( $statements );
+
+		// Valid if the index either precedes or succeeds a statement with the same property
+		return $index > 0 && $index <= $count && $statements[$index - 1]->getPropertyId()->equals( $id )
+			|| $index >= 0 && $index < $count && $statements[$index]->getPropertyId()->equals( $id );
+	}
+
+	/**
+	 * @param Statement[] $statements
+	 * @param int $index
+	 *
+	 * @return bool
+	 */
+	private function isGroupBorder( array $statements, $index ) {
+		// First and last possible position is always a border
+		return $index <= 0
+			|| $index >= count( $statements )
+			|| !$statements[$index - 1]->getPropertyId()->equals( $statements[$index]->getPropertyId() );
+	}
+
+}
--- a/tests/unit/Statement/StatementListChangerTest.php
+++ b/tests/unit/Statement/StatementListChangerTest.php
@@ -1,0 +1,267 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Tests\Statement;
+
+use PHPUnit_Framework_TestCase;
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Services\Statement\StatementListChanger;
+use Wikibase\DataModel\Snak\PropertyNoValueSnak;
+use Wikibase\DataModel\Statement\Statement;
+use Wikibase\DataModel\Statement\StatementList;
+
+/**
+ * @covers Wikibase\DataModel\Services\Statement\StatementListChanger
+ *
+ * @license GPL-2.0+
+ * @author Thiemo Mättig
+ */
+class StatementListChangerTest extends PHPUnit_Framework_TestCase {
+
+	public function groupByPropertyIdProvider() {
+		return [
+			[
+				[],
+				[]
+			],
+			[
+				[ 'P1$a' ],
+				[ 'P1$a' ]
+			],
+			[
+				[ 'P1$a', 'P2$b', 'P1$c', 'P2$d' ],
+				[ 'P1$a', 'P1$c', 'P2$b', 'P2$d' ]
+			],
+			[
+				[ 'P1$a', 'P1$b', 'P2$c', 'P3$d', 'P1$e', 'P2$f' ],
+				[ 'P1$a', 'P1$b', 'P1$e', 'P2$c', 'P2$f', 'P3$d' ]
+			],
+		];
+	}
+
+	/**
+	 * @dataProvider groupByPropertyIdProvider
+	 */
+	public function testGroupByPropertyId( array $guids, array $expectedGuids ) {
+		$statementList = $this->newStatementList( $guids );
+
+		$instance = new StatementListChanger( $statementList );
+		$instance->groupByProperty();
+
+		$this->assertGuids( $expectedGuids, $statementList );
+	}
+
+	public function addToGroupProvider() {
+		return [
+			'add to empty list' => [
+				[],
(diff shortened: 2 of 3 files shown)


src/Statement/StatementListChanger.php
```diff
@@ -0,0 +1,154 @@
...
+use Wikibase\DataModel\Statement\Statement;
+use Wikibase\DataModel\Statement\StatementList;
+
+/**
+ * A collection of functions to manipulate StatementList objects.
+ *
+ * @license GPL-2.0+
+ * @author Thiemo Mättig
+ */
+class StatementListChanger {
```
brightbyte, line 17: It feels to me like this should not be a service, but rather a builder or cursor style object: it should wrap a StatementList it gets in the constructor, and offer specialized operations on that StatementList. Perhaps the name should indicate what kind of extra operations these are.

thiemowmde: I'm very open for suggestions.

I definitely do not want this to be a set of static methods. This is something I would have done a few years ago. I learned. :-)

A builder does not work because I definitely want this to manipulate the original object, not create new ones.

Passing the StatementList via the constructor instead of each method ~~does not change much, on the one hand. On the other hand it makes it much harder to replace this implementation with an other one. You can not simply make this an interface and inject an other implementation. You do not know the StatementList object on construction time. This approach would need a factory.~~

~~In conclusion I still think my current approach is the best.~~

> the name should indicate what kind of extra operations these are.

What do you mean with "extra operations"? What is not obvious enough from the method names?

thiemowmde:
Pinging @Benestar for the sake of completeness, because he spend a lot of time working on this, but we never managed to use one of his approaches. See https://github.com/wmde/WikibaseDataModel/pull/479 and https://github.com/wmde/WikibaseDataModel/pull/507, to name the most relevant ones.

JeroenDeDauw:
I'd put `clear` in `StatementList`. While `StatementList` states "Does not provide complex modification functionality.", `clear` is hardly complex.

If you change `groupByProperty` to a `getGroupedByProperty`, then it also seems to fit into `StatementList`.

That just leaves the group related stuff. I remember that how to best implement this functionality was an old open question already two years ago, at which point there was some investigation into introducing some kind of StatementGroup class. ... found it :)

sg

Start of that thread: https://lists.wikimedia.org/pipermail/wikidata-tech/2014-October/000660.html

Can't find anything about the discussion I mentioned in that mail though; too ancient history... Anyway, something like StatementGroup is what I'd look into for implementing such behaviour.

thiemowmde:
We had a longer discussion about `clear` methods a while ago. My main argument was and still is: Wikibase is all about stable identifiers. It should not be possible to "re-purpose" existing Entities by cleaning and replacing it's contents. Such code should not exist. What we do with Entities is to create them, add, edit and remove individual elements. But we never replace or remove entire sets of Labels, Descriptions, Aliases, Fingerprints, SiteLinks or Statements. The only reason to do this is when test code is written in a way that it reuses existing Entities (which should not be done), or as a temporary workaround for something.

At the moment there is not a single `clear` in the entire DataModel because of this.

Personally I do not like the idea of replacing the wonderful trivial StatementList with something more complex, because this complexity is almost never needed.

JeroenDeDauw:
I see your point about clearing entities being weird. That does not automatically mean having such a method in StatementList, which does not know about entities, is also weird, or as weird. [*] I'm wondering what the clear method being added to this changer class is for now (this is a question). And I'm suspecting that if it being in StatementList is weird, that it being in StatementListChanger is going to be just as weird, plus the feature envy.

[*] Having such a method in StatementList and that being accessible in entities does not mean we think it often makes sense for people to use this. Something making sense depends on the context, and should typically be judged there. We can't really prevent users of code doing insensible things with it. For clearing statement lists, this can be done regardless of having the clear method, as this changer class demonstrates.

## 

> Personally I do not like the idea of replacing the wonderful trivial StatementList with something more complex, because this complexity is almost never needed.

I guess you are talking about the StatementGroup stuff here. Is that correct? If so, I'm not proposing to replace StatementList. It does its job well, and I agree with you that adding something like the group stuff to it increases complexity while most consumers of this code don't care about this functionality. Hence the suggestion for a new thing that holds the group related code.

thiemowmde:
> […] being in StatementListChanger is going to be just as weird […]

Indeed. Can't argue with that. ;-) The only difference is that it is in an other component (Services instead of DataModel itself) and therefore less visible, which makes [what I just explained above](https://github.com/wmde/WikibaseDataModelServices/pull/157#issuecomment-260408559) so much more obvious. For example, nobody will start using this in test setups.

**PS:** I'm fine with removing `clear` entirely, if it helps getting approval.

> I guess you are talking about the StatementGroup stuff here.

Yes.

> Hence the suggestion for a new thing that holds the group related code.

Isn't this what I'm doing here? A new thing that knows how statements are grouped by property? What do you suggest instead?

> If you change groupByProperty to a getGroupedByProperty […]

Turning the code I'm proposing here into getters would not help much. We are usually passing Entities around, not StatementLists. We would need to call `StatementListHolder::setStatements`, which is deprecated because of what I explained above, and the main reason why I came up with a "changer" that just does what needs to be done.

JeroenDeDauw:
> For example, nobody will start using this in test setups.

What kind of bad test code do you image would be written if this method was readily available? If I'm not mistaken it was used in some really shitty tests, though those where not shitty because of that method, and I really hope no one is still writing tests in that manner.

> Isn't this what I'm doing here? A new thing that knows how statements are grouped by property? What do you suggest instead?

I was thinking of an approach different in two ways:
- A class responsible only for the grouping stuff (so you could have a StatementListGrouper if you only want this part)
- A class that holds data and associated behavior. So for instance an actual StatementGroup and StatementGroupList. Perhaps exposed in entities, perhaps not. Perhaps used by them internally, perhaps not.

> Turning the code I'm proposing here into getters would not help much. We are usually passing Entities around, not StatementLists. We would need to call StatementListHolder::setStatements

Thanks for explaining, I was not considering you'd want to use this for modifying entities.

> We would need to call StatementListHolder::setStatements, which is deprecated because of what I explained above

What explanation are you referring to? Is the explanation that you don't want people to be able to clear the statements by using setStatements?

I'm very dubious about this whole Holder stuff. In the scenario you where thinking off, what would the structure of the code be? Something like this?

``` php
function doSomeStuff( StatementListHolder $holder ) {
    // stuff?
    $this->statementChanger->groupByProperty( $holder->getStatements() );
    // stuff?
}
```

What's the problem with

``` php
function doSomeStuff( StatementListHolder $holder ) {
    // stuff?
    $holder->setStatements( $holder->getStatements()->getGroupedByProperty() );
    // stuff?
}
```

While that still has the weird Holder stuff, it's clearer what's going on here (in the former case it's not obvious the $holder ends up with a new state), and the property grouping code does not need to suffer from feature envy.

thiemowmde:
``` php
$changer = new StatementListChanger( $entity->getStatements() );
$changer->groupByProperty();
```

Done. No setter. No "holder".

JeroenDeDauw:
I was asking specifically about the holder scenario that you brought up. You have not answered my questions about that, so now I still don't know what's going on there. I went through the trouble of outlining stuff about the grouper approach and am a disappointed you also did not reply to that.

As for the code you provided now, I certainly agree it is more straightforward. That is due to the lack of the holder, not because of the StatementListChanger. My concern about the readability remains, though only partially. You can still do this instead, though the need is no longer as big (readability wise).

``` php
$entity->setStatements( $entity->getStatements()->getGroupedByProperty() );
```

However in the way you wrote that code, another issue which I was wondering about manifests itself. If you construct the StatementListChanger directly, you have static binding on it and its methods. It's not really different than calling a bunch of namespaced global functions, apart from it superficially looking OO. You can inject the StatementListChanger, though then the readability (which is already not good) suffers.

thiemowmde:
I don't understand what you want to know.

You small example looks nice, but does have the same problem as almost every other possible approach (including what you wrote about "groupers"): it calls a setter that should not be there, and is already deprecated. I don't know what else to say about this.

I went through multiple stages with the code in this patch:
- Methods that don't manipulate the original object but return new ones. We already tried this multiple times (see the linked patches above). It never worked out.
- A set of static methods? Sure, this is possible, but I really want to avoid this.
- No constructor, but have a `$statementList` in every method call? I like this because it makes it possible to inject the service via an interface, and replace the implementation with an other one. But what "other one" would that be? This will never happen.
- So I made it stateful, as it is now.

thiemowmde:
This is still waiting for a review. We want to get rid of the overengineered cruft that is ByPropertyIdArray for a looong, looong time now, and I still believe this here is a wonderful solution. @manicki maybe?

src/Statement/StatementListChanger.php
brightbyte: The method names are fine, but the class name is very generic. "Changer" could be anything. The methods mainly relate to groups, so maybe the class should have "Groups" in the name.  But I can't think of anything nice offhand. So whatever.

src/Statement/StatementListChanger.php
```diff
@@ -0,0 +1,169 @@
...
+
+	public function __construct( StatementList $statementList ) {
+		$this->statementList = $statementList;
+	}
+
+	/**
+	 * Removes all statements from the $statementList object given in the constructor.
+	 */
+	public function clear() {
+		foreach ( $this->statementList->toArray() as $statement ) {
```
brightbyte: It seems to me like clear() should really be part of the container object. And should probably have a much more simple implementations...

src/Statement/StatementListChanger.php
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Makes sure all statements with the same property are next to each other (forming a group),
+	 * and reorders them if necessary. The position of the group in the list is determined by the
+	 * first statement with the same property.
+	 */
+	public function groupByProperty() {
+		$byId = [];
+
+		foreach ( $this->statementList->toArray() as $statement ) {
```
brightbyte, line 36: why is toArray needed here?

src/Statement/StatementListChanger.php
```diff
@@ -0,0 +1,169 @@
...
+			foreach ( $statements as $statement ) {
+				$this->statementList->addStatement( $statement );
+			}
+		}
+	}
+
+	/**
+	 * @param Statement $newStatement
+	 * @param int|null $index An absolute index in the list. If the index is not next to a statement
+	 *  with the same property, the closest possible position is used instead. Default is null,
```
brightbyte: what does "possible" mean here? Does "closest" check in both directions?

src/Statement/StatementListChanger.php
```diff
@@ -0,0 +1,169 @@
...
+ * @author Thiemo Mättig
+ */
+class StatementListChanger {
+
+	/**
+	 * @var StatementList
+	 */
+	private $statementList;
+
+	public function __construct( StatementList $statementList ) {
```
brightbyte, line 24: Thanks!

src/Statement/StatementListChanger.php
```diff
@@ -0,0 +1,169 @@
...
+		if ( $index === null ) {
+			$index = $this->getLastIndexWithinGroup( $statements, $id );
+		} else {
+			// Limit search range to avoid looping non-existing positions
+			$validIndex = min( max( 0, $index ), count( $statements ) );
+			$index = $this->getClosestIndexWithinGroup( $statements, $id, $validIndex );
+			if ( $index === null ) {
+				$index = $this->getClosestIndexAtGroupBorder( $statements, $validIndex );
+			}
+		}
```
brightbyte, line 83: *sigh* This makes it really obvious that we should model groups directly as nested arrays, and get rid of the whole index crap completely.

src/Statement/StatementListChanger.php
```diff
@@ -0,0 +1,169 @@
...
+	 * @param Statement[] $statements
+	 * @param PropertyId $id
+	 * @param int $index
+	 *
+	 * @return bool
+	 */
+	private function isWithinGroup( array $statements, PropertyId $id, $index ) {
+		$count = count( $statements );
+
+		// Valid if the index either prepends ot appends a statement with the same property
```
brightbyte: "precedes or succeeds"?

tests/unit/Statement/StatementListChangerTest.php
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * @license GPL-2.0+
+ * @author Thiemo Mättig
+ */
+class StatementListChangerTest extends PHPUnit_Framework_TestCase {
+
+	public function testConstructor() {
+		$instance = new StatementListChanger( new StatementList() );
+
+		$this->assertInstanceOf( StatementListChanger::class, $instance );
```
brightbyte: why do we need this test?

brightbyte (commented):
Some questions, but looks fine overall.

src/Statement/StatementListChanger.php
thiemowmde: We already have a `StatementGrouper` interface with multiple implementations (including a `ByPropertyIdStatementGrouper`), as well as a `ByPropertyIdArray` (that's what we need to get rid of) and a `ByPropertyIdGrouper`. I wanted this here to be different from these.

This is a layer of code that works with a StatementList while being aware of the concept of "grouping by property ID". ByPropertyIdAwareStatementListChanger?

src/Statement/StatementListChanger.php
thiemowmde: It is optional because StatementList itself is Traversable. But leaving the `->toArray()` out will require an additional `/** @var Statement $statement */` for type safety, and in most places we decided to go with `->toArray()` instead.

src/Statement/StatementListChanger.php
thiemowmde: Yes, both directions. This only matters if the list was not grouped by property ID before.

src/Statement/StatementListChanger.php
thiemowmde: Not sure what you mean with "crap". Can you please avoid using such offensive language?

@Benestar suggested to replace our naive StatementList with one that's aware of PropertyId groups [a long time ago](https://github.com/wmde/WikibaseDataModel/pull/507). I'm not opposing such a change.

I still believe the code proposed here is a helpful step in a longer migration path that might end with replacing StatementList.

Note that the code I'm proposing here does have a very specific advantage: It allows us to add new statements where they belong, without the need to reorder existing statement lists. Not that this is of much relevance, because users usually do not see diffs on the JSON level. But still. This smooth migration is not possible with a StatementList that enforces grouping on construction time.

tests/unit/Statement/StatementListChangerTest.php
thiemowmde: TDD. This is one of the first tests I wrote. It became obsolete later because all other tests are covering the same feature. I just forgot to remove it.

src/Statement/StatementListChanger.php
brightbyte: My comment was not a complaint about this patch. It's just one more place where it becomes obvious that insertion by index into a grouped list is broken by design. It's just bad interface design. And that bad design (aka "crap") is my fault; I remember the session with Denny where we designed this interface. So, *sigh*.

thiemowmde: One of the proposals we discussed had two numeric indexes: one for a relative position in the group, and one for moving the group to an other position. I found this harder to use than the current absolute index.

I'm still very happy with the super-trivial StatementList we have. Only very few places in our application need statements to be grouped, and the requirements are always a bit different. A "baked in" grouping on the DataModel level may help in some cases, but will not solve all use-cases, and will make others more complicated.

JeroenDeDauw:
No changes since 2016. Please reopen if needed for current goals

JeroenDeDauw:
@thiemowmde it is cool you put this effort into getting rid of `ByPropertyIdArray` and it is too bad this stalled.

I looked at this now for 15ish minutes (which means I did not read all comments) and these are my thoughts:

* To me it makes more sense to put this kind of code in WikibaseDataModel. Having gone though this component I spotted some other things I now think belong in there. Keeping all the logic operations on the data structures gives you an Anemic Domain model like we currently have.
* Having a StatementGroup class to represent the groups and host the code dealing with stuff within a group is still a good idea to me. It seems like no comments where made on that suggestion.
* StatementGroup is likely a good place to start to put this code and then the remainder likely fits in StatementList.

Edit: there is a PR relevant to the StatementGroup topic here: https://github.com/wmde/WikibaseDataModel/pull/507

JeroenDeDauw (changes requested):
I think responsibilities are misplaced as per previous comment

JeroenDeDauw:
Let's do a fresh PR that is more focused and does not contain no longer relevant discussion that scares everyone away if someone want to actually create that.

thiemowmde:
This is not meant as "the one" solution to all the year-long "statement (re)ordering" discussions, but merely as a way forward to get rid of ByPropertyIdArray and it's scary capabilities to move statement groups around. Why should I open a new pull request with the same code if it is already here? How are the discussion not relevant?

The concern I raised in https://github.com/wmde/WikibaseDataModel/pull/507#issuecomment-135338904 still holds.

PR #159, opened 2016-11-08
Release 3.7

jakobw:
Some patches outside of this repo depend on `UnknownForeignRepositoryException` so it would be nice to have this released.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -6,6 +6,7 @@
 * Added `EntityIdPager` from `Wikibase\Repo\Store`.
 * Added services for foreign entities:
   * `DispatchingEntityLookup`
+  * `PrefixMappingEntityIdParser`
   * `PrefixMappingEntityIdParserFactory`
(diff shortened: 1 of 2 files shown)


manicki:
`PrefixMappingEntityIdParser` has been also added. I am not sure if want to explicitly mention it, or is it OK to only mentioned the factory class.

jakobw:
Added it!

JeroenDeDauw:
If the thing is package public, it should be mentioned. If it's package private, it should not be mentioned, though then its good to mention in the class doc that it is indeed package private (instead of having a since tag)

manicki:
~~-1
Please note that per https://phabricator.wikimedia.org/T150328 `DispatchingEntityLookup` should be adjusted. This should happen before releasing a new version of this component.~~

manicki:
`DispatchingEntityLookup` is now ready to be released.

manicki:
Added sentence saying that we raised DM version requirement

PR #160, opened 2016-11-09
Make DispatchingEntityLookup not throw exceptions on ids from the unknown repository

manicki:
Entities from unknown repositories should generally be treated as missing entities, not result in exceptions.

See also https://phabricator.wikimedia.org/T150328


brightbyte:
+1 from me. I'm leaving this here for a day, will merge it tomorrow if there are no objections. Poke me.

thiemowmde:
:+1: Makes sense to me. Code looks fine.

PR #161, opened 2016-11-11
Clarify return values of DispatchingEntityLookup methods

manicki:


thiemowmde:
I would add this directly to the `@return` tag. :+1: otherwise.

manicki:
Changed

PR #164, opened 2016-12-05
Fix spacing

manicki:
I noticed this missing space staring at me this morning. Don't we have some sniff for detecting such fixing in place somewhere?


thiemowmde:
The sniff that does report this is part of the MediaWiki rule set. See #165.

PR #166, opened 2016-12-13
Inline trivial returnCallback as closure

thiemowmde:
This reduces duplicate code (by turning multiple `if` into a `switch`) and makes a public function private (by turning it into an inlined closure).


Ladsgroup:
It's good but you also changed several if conditions to a switch condition which is good but it's not mentioned in the commit message or title. Please add it too.

thiemowmde:
"Reduces duplicate code" by utilizing a `switch`. I don't think it's helpful to repeat all implementation details in the commit message. It should explain the "why".

Ladsgroup:
Okay and by "makes a public function private." do you mean the closure?

thiemowmde:
Yes. I updated the commit message.

Ladsgroup:
Thanks

PR #173, opened 2017-02-27
Avoid calling getFingerprint where not necesarry

thiemowmde:
This goes along with https://github.com/wmde/WikibaseDataModel/pull/717.

### DIFF
--- a/src/Diff/Internal/AliasGroupListPatcher.php
+++ b/src/Diff/Internal/AliasGroupListPatcher.php
@@ -7,7 +7,6 @@ use Diff\DiffOp\DiffOp;
 use Diff\DiffOp\DiffOpAdd;
 use Diff\DiffOp\DiffOpChange;
 use Diff\DiffOp\DiffOpRemove;
-use Diff\Patcher\MapPatcher;
 use Diff\Patcher\PatcherException;
 use Wikibase\DataModel\Term\AliasGroupList;
 
@@ -37,7 +36,7 @@ class AliasGroupListPatcher {
 	}
 
 	/**
-	 * @see MapPatcher
+	 * @see Diff\Patcher\MapPatcher
 	 *
 	 * @param AliasGroupList $groups
 	 * @param string $lang
--- a/src/Diff/ItemDiffer.php
+++ b/src/Diff/ItemDiffer.php
@@ -75,9 +75,9 @@ class ItemDiffer implements EntityDifferStrategy {
 	private function toDiffArray( Item $item ) {
 		$array = [];
 
-		$array['aliases'] = $item->getFingerprint()->getAliasGroups()->toTextArray();
-		$array['label'] = $item->getFingerprint()->getLabels()->toTextArray();
-		$array['description'] = $item->getFingerprint()->getDescriptions()->toTextArray();
+		$array['aliases'] = $item->getAliasGroups()->toTextArray();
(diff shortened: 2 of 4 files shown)


src/Diff/Internal/AliasGroupListPatcher.php
```diff
@@ -37,7 +36,7 @@ public function patchAliasGroupList( AliasGroupList $groups, Diff $patch ) {
 	}
 
 	/**
-	 * @see MapPatcher
+	 * @see Diff\Patcher\MapPatcher
```
manicki, line 39: Why this change?

thiemowmde: My PHPStorm does not understand this if it is not a full class name.

manicki: oh, I didn't notice there is no `use Diff\Patcher\MapPatcher` on top of the file (which makes the tool not able to understand the @see thing, which makes sense).

manicki: meh, I am still half-awake. Why removing the import here then? I don't know all stuff PHP does internally but having the "use" statement doesn't mean the class is loaded, it only gets loaded once used in code (ie. not in the documentation). So this import wouldn't make any change regarding performance, memory usage etc as far as I get it (but maybe I am wrong).

I am not opposing to this change here. I am just a bit surprised as I've seen multiple places in Wikibase which import things just to have a simpler reference in the phpdoc to the class/its method, so I thought this is a general practice.

manicki:
+1
Looks good, I am going to merge this.
Just giving @thiemowmde few minutes to read my ridiculous comment before I hit the button.

src/Diff/Internal/AliasGroupListPatcher.php
thiemowmde: The stuff in the `use` section [is nothing but a string map](http://maettig.com/1457600220) and not loading anything. I'm aware of that.

What you have seen are `@return`, `@param` and `@var` tags. This here is a `@see`. PHPStorm does not consider the `use` section when resolving `@see` tags (why should it?). `@see` only works with local methods, classes that are in the same namespace, or fully qualified class names.

manicki: I am confused by this evil tool. So my Phpstorm recognizes MapPatcher in this `@see` link when not using FQN. I thought it only happens with "use" import but apparently it is smart enough even when "use" is not there.
It seems like a tool-specific thing, I don't care then.

Per this part of your comment:
> What you have seen are `@return`, `@param` and `@var` tags. This here is a `@see`.

So what you're saying is there is a rule that imported classes are considered while resolving `@return`, `@param` and `@var` comment but not in case of a `@see`? Fine if there is something like this somewhere but it seems like a good reason to be confused about :)

manicki:
PR is fine. @manicki can take his fascinating monologue on being confused by IDE behaviour to some other place.

src/Diff/Internal/AliasGroupListPatcher.php
thiemowmde: `@return`, `@param` and `@var` describe actual behaviour of the code, while `@see` is more like an `<a href>` to guide the reader. Considering this the difference makes sense to me.

PR #176, opened 2017-04-20
Introduce LabelLookup interface

thiemowmde:
Discussed with the team during one of the past engineering times. @brightbyte gave the most relevant input, which says that the LabelDescriptionLookup interface combining labels and descriptions is indeed very useful, and should not be split into two separate interfaces.

Think of it this way: A description alone is not useful. You might want to render an `<a>`nchor using the label as the visible text and the description as `title="…"`.

But when you check the usages of the two `getLabel` and `getDescription` methods you realize we do have a **lot** of use-cases that use a label but **no** description. This new interface is not going to change anything, because all implementations will stay the same (for now). But it helps us cleaning up our code by using interfaces that are as narrow as possible.

[Bug: T163538](https://phabricator.wikimedia.org/T163538)

### DIFF
--- a/src/Lookup/LabelLookup.php
+++ b/src/Lookup/LabelLookup.php
@@ -6,22 +6,19 @@ use Wikibase\DataModel\Entity\EntityId;
 use Wikibase\DataModel\Term\Term;
 
 /**
- * @since 1.1
+ * @since 3.10
  *
- * @license GPL-2.0+
- * @author Katie Filbert < aude.wiki@gmail.com >
- * @author Marius Hoch < hoo@online.de >
+ * @license GPL-2.0-or-later
+ * @author Thiemo Kreuz
  */
-interface LabelDescriptionLookup extends LabelLookup {
+interface LabelLookup {
 
 	/**
-	 * @since 2.0
-	 *
 	 * @param EntityId $entityId
 	 *
 	 * @throws LabelDescriptionLookupException
 	 * @return Term|null
 	 */
-	public function getDescription( EntityId $entityId );
+	public function getLabel( EntityId $entityId );
 
 }
--- a/tests/unit/EntityId/EntityIdLabelFormatterTest.php
+++ b/tests/unit/EntityId/EntityIdLabelFormatterTest.php
@@ -7,8 +7,8 @@ use Wikibase\DataModel\Entity\EntityId;
 use Wikibase\DataModel\Entity\ItemId;
 use Wikibase\DataModel\Entity\PropertyId;
 use Wikibase\DataModel\Services\EntityId\EntityIdLabelFormatter;
-use Wikibase\DataModel\Services\Lookup\LabelDescriptionLookup;
 use Wikibase\DataModel\Services\Lookup\LabelDescriptionLookupException;
+use Wikibase\DataModel\Services\Lookup\LabelLookup;
 use Wikibase\DataModel\Term\Term;
 
 /**
@@ -46,8 +46,7 @@ class EntityIdLabelFormatterTest extends PHPUnit_Framework_TestCase {
 	 * @param string $expectedString
 	 */
 	public function testParseWithValidArguments( EntityId $entityId, $languageCode, $expectedString ) {
-		$labelDescriptionLookup = $this->getLabelDescriptionLookup( $languageCode );
-		$formatter = new EntityIdLabelFormatter( $labelDescriptionLookup );
+		$formatter = new EntityIdLabelFormatter( $this->getLabelLookup( $languageCode ) );
 
 		$formattedValue = $formatter->formatEntityId( $entityId );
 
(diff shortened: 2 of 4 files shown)


src/Lookup/LabelLookup.php
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * @since 3.9
+ *
+ * @license GPL-2.0+
+ * @author Thiemo Mättig
+ */
+interface LabelLookup {
+
+	/**
+	 * @since 2.0
```
JeroenDeDauw: nope

JeroenDeDauw: Actually I'm not convinced it makes sense to have a tag at all on method level in a codebase where you have sane interfaces. And given there is just one method here...

tests/unit/EntityId/EntityIdLabelFormatterTest.php
```diff
@@ -58,14 +57,14 @@ public function testParseWithValidArguments( EntityId $entityId, $languageCode,
 	/**
 	 * @param string $languageCode
 	 *
-	 * @return LabelDescriptionLookup
+	 * @return LabelLookup
 	 */
-	private function getLabelDescriptionLookup( $languageCode ) {
-		$labelDescriptionLookup = $this->getMockBuilder( 'Wikibase\DataModel\Services\Lookup\LabelDescriptionLookup' )
+	private function getLabelLookup( $languageCode ) {
+		$labelLookup = $this->getMockBuilder( 'Wikibase\DataModel\Services\Lookup\LabelLookup' )
```
JeroenDeDauw: I'd have taken the opportunity to switch this to ::class, just saying

JeroenDeDauw (approved):
Reminds me of https://github.com/JeroenDeDauw/TermStore/blob/master/src/LabelLookup.php

brightbyte:
My 2¢: we can do this, but if EntityIdLabelFormatter is the only reason to introduce the new interface, I don't think it's worth the trouble. In fact, I think that EntityIdHtmlLinkFormatter, which extends EntityIdLabelFormatter, should indeed use a LabelDescriptionLookup and also make the description accessible - perhaps in the title attribute.

thiemowmde:
This interface is not meant to be used in implementations that output HTML. These should indeed have access to both labels and description. This interface is meant for all *other* non-HTML use cases where only a label is needed.

JeroenDeDauw:
> worth the trouble

the cost of having such an interface is almost 0

src/Lookup/LabelLookup.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @license GPL-2.0+
+ * @author Thiemo Mättig
+ */
+interface LabelLookup {
+
+	/**
+	 * @param EntityId $entityId
+	 *
+	 * @throws LabelDescriptionLookupException
```
mariushoch, line 19: This is nitpicking, but this exception is not totally accurate here :/

JeroenDeDauw: Agreed

src/Lookup/LabelLookup.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Lookup;
+
+use Wikibase\DataModel\Entity\EntityId;
+use Wikibase\DataModel\Term\Term;
+
+/**
+ * @since 3.9
```
JeroenDeDauw: This does not match what is in the README

src/Lookup/LabelLookup.php
```diff
@@ -0,0 +1,24 @@
...
+ * @license GPL-2.0+
+ * @author Thiemo Mättig
+ */
+interface LabelLookup {
+
+	/**
+	 * @param EntityId $entityId
+	 *
+	 * @throws LabelDescriptionLookupException
+	 * @return Term|null
```
JeroenDeDauw, line 20: Not an issue with this PR, just an FYI: looking at this code after having been a run away'd from Wikibase for some time now, I find it weird this type is Term and not Label.

JeroenDeDauw:
It is weird work is done and then sits here for a year. Either the work done was not useful or useful work is being wasted. Either way, not good.

src/Lookup/LabelLookup.php
thiemowmde: There is no Label class, but a common Term class for both labels and descriptions.

src/Lookup/LabelLookup.php
thiemowmde: I suggest to rename this exception to `TermLookupException` or something similar. Can we please do this later, in a separate patch?

thiemowmde:
I think the problem with proposals like https://github.com/DataValues/Common/pull/35 and this one here is that they do not "magically" solve an issue just because a smaller, more cleanly designed interface is proposed in a subcomponent. The benefit only becomes obvious when the thing is put to use and proved it's right to exist over time. But this can never happen without it being reviewed, merged, released, and actually used first.

In this particular case I found the issue obvious: There are more than 100 type hints against the LabelDescriptionLookup interface in the Wikibase code base (60 in production), but many, many of them don't care the slightest about descriptions. They **are** LabelLookups. If we could actually name them like this, we could have so much more clarity.

I always found the need to somehow "proof" the usefulness of an idea first – outside of the context where it's usefulness would just be obvious – endlessly frustrating. Sure, I could prepare a patch in Wikibase that replaces **all** LabelDescriptionLookup that can be replaced with LabelLookup (because replacing only one would not proof anything). But if not even a minimal proposal like the one in this pull request is appreciated, how would a 1000 line patch in Wikibase help?

src/Lookup/LabelLookup.php
mariushoch: Sure thing :)

PR #178, opened 2017-06-01
Make use of the …::class feature

thiemowmde:


Ladsgroup:
Travis failure is not related but needs to be addressed ASAP.

PR #180, opened 2017-06-01
Add removed PHPCS sniffs back

thiemowmde:
In #179 I unintentionally removed a lot of rules. I just want to go back to how it was before in this code repository, without judging if a rule makes sense or not. We might need to remove some that are to strict. We might include one or two in the wikibase-codesniffer library. This is for later to discuss.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: php
 
+dist: trusty
+
 env:
   - THENEEDFORTHIS=FAIL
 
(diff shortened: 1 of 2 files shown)


Ladsgroup:
Travis is not happy about it. https://travis-ci.org/wmde/WikibaseDataModelServices/jobs/238357311

Ladsgroup (changes requested):
Please fix CS errors :)

thiemowmde:
Done. I decided to remove `Generic.NamingConventions.CamelCapsFunctionName` because it is a duplicate of `PSR1.Methods.CamelCapsMethodNameSniff`, and I don't want to repeat the required exclusion.

Ladsgroup (approved):

PR #182, opened 2017-06-23
Move EntityIdComposer from Wikibase to DataModelServices

Ladsgroup:


src/EntityId/EntityIdComposer.php
```diff
@@ -0,0 +1,69 @@
...
+use Wikibase\DataModel\Entity\EntityId;
+
+/**
+ * Constructs EntityId objects from entity type identifiers and unique parts of entity ID
+ * serializations. The unique part is typically the numeric part of an entity ID, excluding the
+ * static part that's the same for all IDs of that type.
+ *
+ * Meant to be the counterpart for @see Int32EntityId::getNumericId, as well as an extensible
+ * replacement for @see LegacyIdInterpreter::newIdFromTypeAndNumber.
+ *
```
thiemowmde, line 16: Can you add a `@since` tag?

tests/unit/EntityId/EntityIdComposerTest.php
```diff
@@ -0,0 +1,99 @@
...
+use PHPUnit_Framework_TestCase;
+use UnexpectedValueException;
+use Wikibase\DataModel\Entity\EntityId;
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Services\EntityId\EntityIdComposer;
+
+/**
+ * @covers Wikibase\DataModel\Services\EntityId\EntityIdComposer
+ *
+ * @group Wikibase
```
thiemowmde: Please remove, I think we don't use `@group` tags in this component.

thiemowmde (changes requested):
It even said "please move me", so yay! Just a few documentation details.

tests/unit/EntityId/EntityIdComposerTest.php
Ladsgroup: Done

src/EntityId/EntityIdComposer.php
Ladsgroup: Done

PR #184, opened 2017-08-02
Remove auto-generated @param-only PHPDoc blocks that add zero information

thiemowmde:
This patch removes auto-generated PHPDoc blocks that contain nothing but a sequence of `@param` lines that *literally* repeat what the function header below contains. These blocks add zero additional information, but clutter the code and need to be maintained. This patch does not touch PHPDoc blocks when one of the `@param` lines contains an actual comment behind the variable name, or when one of the types is not a class name.

See https://gerrit.wikimedia.org/r/360814 for the same patch in the Wikibase.git code base.

### DIFF
--- a/src/Entity/PropertyDataTypeMatcher.php
+++ b/src/Entity/PropertyDataTypeMatcher.php
@@ -21,9 +21,6 @@ class PropertyDataTypeMatcher {
 	 */
 	private $propertyDataTypeLookup;
 
-	/**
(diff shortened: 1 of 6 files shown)


Ladsgroup:
Can you rebase?

thiemowmde:
Something was failing for a reason I did not understood. I hope it disappears with the rebase I did now.

PR #186, opened 2017-09-11
BUG: Statements are not merged when StatementList is patched

bekh6ex:
While I was trying to figure out how I should patch Lexeme Forms I tried to understand how the things are for statements and that is what I discovered:
Individual statements are never patched. Each statement just gets replaced with the new one. As I see it it is a problem, taking into account the possibility to "edit old revisions" (I mean that client sends `baseRevId`) it might be quiet harmful if you edit qualifiers or references, because those might be silently removed from the latest revision.

### DIFF
--- a/tests/unit/Diff/StatementListDifferPatcherTest.php
+++ b/tests/unit/Diff/StatementListDifferPatcherTest.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Tests\Diff;
+
+use Wikibase\DataModel\Services\Diff\StatementListDiffer;
+use Wikibase\DataModel\Services\Diff\StatementListPatcher;
+use Wikibase\DataModel\Snak\PropertyNoValueSnak;
+use Wikibase\DataModel\Snak\SnakList;
+use Wikibase\DataModel\Statement\Statement;
+use Wikibase\DataModel\Statement\StatementList;
+
+class StatementListDifferPatcherTest extends \PHPUnit_Framework_TestCase {
+
+	public function testStatemetsAreMergedWhenPatched() {
+		$guid = 'some-guid';
+		$statementLatest = new Statement(
+			new PropertyNoValueSnak( 1 ),
+			new SnakList( [ new PropertyNoValueSnak( 3 ) ] ),
+			null,
+			$guid
+		);
+		$statementFrom = new Statement(
+			new PropertyNoValueSnak( 1 ),
+			new SnakList( [ new PropertyNoValueSnak( 1 ) ] ),
+			null,
+			$guid
+		);
+		$statementTo = new Statement(
+			new PropertyNoValueSnak( 1 ),
+			new SnakList( [ new PropertyNoValueSnak( 2 ) ] ),
+			null,
+			$guid
+		);
+		$latestStatementList = new StatementList( [ $statementLatest ] );
+
+		$differ = new StatementListDiffer();
+		$patcher = new StatementListPatcher();
+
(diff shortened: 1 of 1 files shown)


tests/unit/Diff/StatementListDifferPatcherTest.php
```diff
@@ -0,0 +1,47 @@
...
+		$differ = new StatementListDiffer();
+		$patcher = new StatementListPatcher();
+
+		$diff = $differ->getDiff( new StatementList( [ $statementFrom ] ), new StatementList( [ $statementTo ] ) );
+		$patcher->patchStatementList( $latestStatementList, $diff );
+
+		$statement = $latestStatementList->getFirstStatementWithGuid( $guid );
+		$this->assertEquals( new PropertyNoValueSnak( 3 ), $statement->getQualifiers()[0] );
+		$this->assertEquals( new PropertyNoValueSnak( 2 ), $statement->getQualifiers()[1] );
+		$this->assertEquals( 2, $statement->getQualifiers()->count() );
```
JeroenDeDauw, line 45: you can use assertCount

tests/unit/Diff/StatementListDifferPatcherTest.php
```diff
@@ -0,0 +1,47 @@
...
+			null,
+			$guid
+		);
+		$latestStatementList = new StatementList( [ $statementLatest ] );
+
+		$differ = new StatementListDiffer();
+		$patcher = new StatementListPatcher();
+
+		$diff = $differ->getDiff( new StatementList( [ $statementFrom ] ), new StatementList( [ $statementTo ] ) );
+		$patcher->patchStatementList( $latestStatementList, $diff );
```
JeroenDeDauw, line 40: This test shows there is an issue either with diffing or with patching. Did you verify the issue is indeed with the latter?

If I wanted to test for the suggested bug I'd have written this test without the differ, and just have a statement and a diff to apply to it. What made you go with the more integration approach?

JeroenDeDauw (commented):
Looks good to me

tests/unit/Diff/StatementListDifferPatcherTest.php
bekh6ex: Because I have no clue what Diff should look like to create it by hands.

tests/unit/Diff/StatementListDifferPatcherTest.php
bekh6ex: Maybe, but the point of this test code is just to present existence of the bug.

tests/unit/Diff/StatementListDifferPatcherTest.php
bekh6ex: And another reason to have it like this is that `Diff`, as far as I understand, currently is not created anywhere except in the Differ, so we might not treat `Diff`'s constructor as a public interface.
In this case this way of testing it makes perfect sense.

JeroenDeDauw: +1

thiemowmde:
Can you please open a Phabricator ticket for this? I can confirm the issue, and would classify it as a bug. It's quite easy to reproduce: Open the same entity in two tabs. Add a qualifier or a reference in one tab. Add an other qualifier or reference in the other tab (with the same property). The additions from the first tab are lost.

JonasKress:
https://phabricator.wikimedia.org/T175836

JeroenDeDauw:
I will close this as it is already tracked by a phabricator ticket. When that is picked up this PR can be reopened or a new one can be created as desired.

PR #187, opened 2018-01-11
Add DisabledEntityTypesEntityLookup

Ladsgroup:
Bug: https://phabricator.wikimedia.org/T178904

### DIFF
--- a/tests/unit/Lookup/DisabledEntityTypesEntityLookupTest.php
+++ b/tests/unit/Lookup/DisabledEntityTypesEntityLookupTest.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Tests\Lookup;
+
+use Wikibase\DataModel\Entity\EntityId;
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Services\Lookup\DisabledEntityTypesEntityLookup;
+use Wikibase\DataModel\Services\Lookup\EntityLookup;
+use Wikibase\DataModel\Services\Lookup\EntityLookupException;
+
+/**
+ * @covers Wikibase\DataModel\Services\Lookup\DisabledEntityTypesEntityLookup
+ *
+ * @license GPL-2.0+
+ * @author Amir Sarabadani
+ */
+class DisabledEntityTypesEntityLookupTest extends \PHPUnit_Framework_TestCase {
+
+	/**
+	 * @return EntityLookup
+	 */
+	private function getEntityLookup() {
+		$entityLookup = $this->getMock( EntityLookup::class );
+
+		$entityLookup->expects( $this->any() )
+			->method( 'hasEntity' )
+			->will( $this->returnValue( true ) );
+
+		$entityLookup->expects( $this->any() )
+			->method( 'getEntity' )
+			->will( $this->returnCallback( function( EntityId $id ) {
+				return $id->getSerialization();
+			} ) );
+
+		return $entityLookup;
+	}
+
+	public function testConstructor() {
+		$lookup = new DisabledEntityTypesEntityLookup( $this->getEntityLookup(), [] );
+		$this->assertInstanceOf(
+			DisabledEntityTypesEntityLookup::class,
+			$lookup
+		);
+	}
+
+	/**
+	 * @expectedException \InvalidArgumentException
+	 */
+	public function testConstructor_exception() {
+		new DisabledEntityTypesEntityLookup( $this->getEntityLookup(), [ 0 ] );
+	}
+
+	public function testHasEntity() {
+		$lookup = new DisabledEntityTypesEntityLookup( $this->getEntityLookup(), [] );
+
+		$this->assertTrue( $lookup->hasEntity( new ItemId( 'Q22' ) ) );
+	}
+
+	public function testGetEntity() {
+		$lookup = new DisabledEntityTypesEntityLookup( $this->getEntityLookup(), [] );
+
+		for ( $i = 1; $i < 6; $i++ ) {
+			$this->assertSame(
+				'Q' . $i,
+				$lookup->getEntity( new ItemId( 'Q' . $i ) )
+			);
+		}
+	}
+
+	public function testGetEntity_exceptionDisabledEntityType() {
+		$lookup = new DisabledEntityTypesEntityLookup( $this->getEntityLookup(), [ 'item' ] );
+		$this->setExpectedException( EntityLookupException::class );
+		$lookup->getEntity( new ItemId( 'Q1' ) );
+	}
+
+}
--- a/src/Lookup/DisabledEntityTypesEntityLookup.php
+++ b/src/Lookup/DisabledEntityTypesEntityLookup.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Lookup;
+
+use Wikibase\DataModel\Entity\EntityDocument;
+use Wikibase\DataModel\Entity\EntityId;
+use Wikimedia\Assert\Assert;
+
+/**
+ * EntityLookup that checks entities have been loaded through it and throws
+ * an exception if the accessing to that entity type is disabled.
+ *
+ * @since 3.9
+ *
+ * @license GPL-2.0+
+ * @author Amir Sarabadani
+ */
+class DisabledEntityTypesEntityLookup implements EntityLookup {
+
+	/**
+	 * @var EntityLookup
+	 */
+	private $entityLookup;
+
+	/**
+	 * @var string[]
+	 */
+	private $disabledEntityTypes = [];
+
+	/**
+	 * @param EntityLookup $entityLookup
+	 * @param string[] $disabledEntityTypes
+	 */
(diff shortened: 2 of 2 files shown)


tests/unit/Lookup/RestrictedEntityLookupTest.php
```diff
@@ -37,7 +38,10 @@ private function getEntityLookup() {
 
 	public function testConstructor() {
 		$lookup = new RestrictedEntityLookup( $this->getEntityLookup(), 1 );
-		$this->assertInstanceOf( RestrictedEntityLookup::class, $lookup );
+		$this->assertInstanceOf(
+			'Wikibase\DataModel\Services\Lookup\RestrictedEntityLookup',
```
JeroenDeDauw: I was about to suggest using ::class and then saw you actually moved away from it. I am confoose.

Ladsgroup: Oh, It was a rebase mistake. Let me fix it.

JeroenDeDauw:
This is a single responsibility violation. The decorator is now doing two different things. Granted, they both restrict access in some way, so I can see how this class might seem like a good place for this new functionality given its name. I'd create a new decorator and consider renaming this class.

tests/unit/Lookup/RestrictedEntityLookupTest.php
```diff
@@ -101,6 +105,13 @@ public function testGetEntity_exception() {
 		$lookup->getEntity( new ItemId( 'Q4' ) );
 	}
 
+	public function testGetEntity_exceptionDisabledEntityType() {
+		$lookup = new RestrictedEntityLookup( $this->getEntityLookup(), 200 );
+		$lookup->setDisabledEntityTypes( [ 'item' ] );
+		$this->setExpectedException( EntityLookupException::class );
```
JeroenDeDauw: This is PHPUnit syntax that is no longer in any supported version. Not an issue with this PR though I do want to raise the question if this component can use a version of PHPUnit that is supported.

thiemowmde: This is very much unrelated to anything going on here. Currently this syntax is used 500+ times in all Wikibase related codebases combined. When we get the chance to switch to newer versions (which will happen soon), this is mostly a trivial string replace (`setExpectedException` → `expectException`, except we test for exception messages, which we barely do).

src/Lookup/RestrictedEntityLookup.php
```diff
@@ -129,4 +140,11 @@ public function entityHasBeenAccessed( EntityId $entityId ) {
 		return array_key_exists( $entityIdSerialization, $this->entitiesAccessed );
 	}
 
+	/**
+	 * @param string[] $disabledEntityTypes
+	 */
+	public function setDisabledEntityTypes( array $disabledEntityTypes ) {
```
thiemowmde: I disagree with the idea to have a setter for this. It introduces an issue I find quite problematic: the behavior of this service can be changed any time by any code that is allowed to use this service. This should be a constructor argument.

When doing so it also becomes a little more obvious that Jeroen's suggestion to introduce an other decorator (and possibly rename this one here) might be a cleaner solution.

thiemowmde (changes requested):
I do agree with Jeroen. However, I don't want to block this route. I am willing to have this feature added to this class. But I feel it is equally easy to introduce a new class for this feature.

src/Lookup/RestrictedEntityLookup.php
Ladsgroup: I wanted to go with constructor but Daniel suggested to go with setter specially because it has a sensible (empty array) default. FWIW, How much does it matter at the end?

Ladsgroup:
If we want to really comply with SRP, We should make two classes, one AccessLimitBasedRestriction and the other one EntityTypeDisabledRestriction (which both should implement an interface e.g. called EntityAccessRestriction) and make the RestrictedEntityLookup get them in array and check. That's the proper way to do it but if you want to implement that, have fun.

src/Lookup/RestrictedEntityLookup.php
thiemowmde: As explained I believe it matters a lot.

thiemowmde:
I don't think such an infrastructure (an array of EntityAccessRestrictions) is needed. The decorator pattern – as I understand it – does not need such an interface. All you need is just an other class that implements EntityLookup, and wraps the original lookup. You can copy the RestrictedEntityLookup and slightly rework it (locally in the client directory, where it is needed). That's not hard in my opinion.

JeroenDeDauw:
Yeah no need for such an interface as far as I can see. What made you suggest it?

tests/unit/Lookup/RestrictedEntityLookupTest.php
manicki: So to answer the original question:
As long as Wikibase supports PHP 5.5 we are bound to PHPUnit 4.8. Relatively soon (although I think rather months than weeks) we might be able to go up to PHP 5.6 and PHPUnit 5. But it is not really great news I think, as when we do it, PHPUnit 5 will already be EOL. Full switch to PHP 7 is going to take some more (hopefully before PHP 7 and PHPUnit 6 are ancient).

So sadly, I think the answer to what @JeroenDeDauw asked is: not really.

manicki:
I've had a brief look, and I agree with @JeroenDeDauw that the separate decorator would be better. We've talked with @Ladsgroup and I believe we share this view, and we clarified how to implement this.

Re setter vs constructor parameter: i guess with the separate decorator this is going to be a redundant question. But as the setter suggestion came from @brightbyte, he could provide some thoughts on this if needed.

brightbyte:
Re setter vs constructor parameter: no strong feelings for this concrete case. As a general rule of thumb, I prefer setters for things that have a sensible default, and reserve constructor params for things that don't. That avoids long parameter lists and pointless dummy values in tests. 

As to making this a separate decorator: it's cleaner, but more overhead. seems overkill to me. but again, no strong preference.

tests/unit/Lookup/DisabledEntityTypesEntityLookupTest.php
```diff
@@ -0,0 +1,76 @@
...
+		$entityLookup->expects( $this->any() )
+			->method( 'getEntity' )
+			->will( $this->returnCallback( function( EntityId $id ) {
+				return $id->getSerialization();
+			} ) );
+
+		return $entityLookup;
+	}
+
+	public function testConstructor() {
```
manicki, line 38: probably this test could even be skipped but whatever (I mean the constructor works as other test prove that).
I don't mind it stay though.

manicki (approved):
+1
Thanks @Ladsgroup for re-doing!
Looks fine for me. As this PR initially got quite some attention, I'll give others some time (but not too much) to have a look. For naming suggestions and stuff.
You might maybe want to change the description of this PR, so in the future possible people discovering will not be confoosed.

JeroenDeDauw:
> I prefer setters for things that have a sensible default, and reserve constructor params for things that don't. That avoids long parameter lists and pointless dummy values in tests.

In my view this is an anti-pattern and it is not something done in the fundraising codebases. If your parameter lists are too long it suggests too many responsibilities. You can see this with this code. You'd have just 2 constructor parameters when you create a new decorator. If you repeat pointless dummy values often in your tests this suggests you write your tests badly and I suggest have a look at how we write integration tests for usecases in the FunrdaisingFrontend (these sometimes deal with a bunch of constructor parameters).

brightbyte:
>  In my view this is an anti-pattern and it is not something done in the fundraising codebases. If your parameter lists are too long it suggests too many responsibilities. 

To me this is a tradeoff. E.g. have an optional logger, or a logging decorator? Have a default batch size, or force the batch size to be explicit? Anyway, any constructor argument that has a default value can just as well have a setter instead. I recommend a pragmatic choice on a case by case basis. If PHP has named parameters like Python, this wouldn't be an issue.

In this case, perhaps the deciding point is consistency: $entityAccessLimit is forced to be given explicitly, so perhaps the list of disabled types should also be explicit. Or make a setter for  $entityAccessLimit.

Also, the bike shed should be purple. Light purple.

mariushoch (approved):
Seems sensible to me.

thiemowmde:
> any constructor argument that has a default value can just as well have a setter instead.

I'm afraid I have to repeat what I already wrote above: No, "any constructor argument" can **not** "as well" be a setter. There is a humongous difference between these two approaches: A constructor argument can only be passed once. A setter can be called any time.

If the class we are talking about is a service that is (ideally) constructed once and then passed to many different users, it would be horrible if users could change the behavior of the service any time they like, and interfere with all other users. I hope I don't need to explain why such a situation would be bad.

To avoid this you would need to lock-down the setter or the object somehow, so the setter can only be called once. But doing so would not fit to the expected behavior of a setter. You could as well use a constructor argument then.

brightbyte:
@thiemo You are right that this kind of setter should only be used in bootstrap code, and it's bad to expose it to application logic. This can easily be avoided by having a service interface that does not include the setter. And/or you call it an initializer instead of a setter, so there is no confusion about expected behavior.

But really, we are no longer talking about the concrete case here. So let's take this somewhere else.

PR #195, opened 2018-04-14
Introduce EntityRetrievingClosestReferencedEntityIdLookup

mariushoch:
Part of https://phabricator.wikimedia.org/T179155

### DIFF
--- a/src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
+++ b/src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
@@ -1,0 +1,249 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Lookup;
+
+use Wikibase\DataModel\Entity\EntityId;
+use Wikibase\DataModel\Entity\EntityIdValue;
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Services\Entity\EntityPrefetcher;
+use Wikibase\DataModel\Snak\PropertyValueSnak;
+use Wikibase\DataModel\Snak\Snak;
+use Wikibase\DataModel\Statement\StatementListProvider;
+
+/**
+ * Service for getting the closest entity (out of a specified set),
+ * from a given starting entity. The starting entity, and the target entities
+ * are (potentially indirectly, via intermediate entities) linked by statements
+ * with a given property ID, pointing from the starting entity to one of the
+ * target entities.
+ *
+ * @since 3.10
+ *
+ * @license GPL-2.0-or-later
+ * @author Marius Hoch
+ */
+class EntityRetrievingClosestReferencedEntityIdLookup implements ReferencedEntityIdLookup {
+
+	/**
+	 * @var EntityLookup
+	 */
+	private $entityLookup;
+
+	/**
+	 * @var EntityPrefetcher
+	 */
+	private $entityPrefetcher;
+
+	/**
+	 * @var int Maximum search depth: Maximum number of intermediate entities to search through.
+	 *  For example 0 means that only the entities immediately referenced will be found.
+	 */
+	private $maxDepth;
+
+	/**
+	 * @var int Maximum number of entities to retrieve.
+	 */
+	private $maxEntityVisits;
+
+	/**
+	 * Map (entity id => true) of already visited entities.
+	 *
+	 * @var bool[]
+	 */
+	private $alreadyVisited = [];
+
+	/**
+	 * @param EntityLookup $entityLookup
+	 * @param EntityPrefetcher $entityPrefetcher
+	 * @param int $maxDepth Maximum search depth: Maximum number of intermediate entities to search through.
+	 *  For example if 0 is given, only the entities immediately referenced will be found.
+	 *  If this limit gets exhausted, a MaxReferenceDepthExhaustedException is thrown.
+	 * @param int $maxEntityVisits Maximum number of entities to retrieve during a lookup.
+	 *  If this limit gets exhausted, a MaxReferencedEntityVisitsExhaustedException is thrown.
+	 */
+	public function __construct(
+		EntityLookup $entityLookup,
+		EntityPrefetcher $entityPrefetcher,
+		$maxDepth,
+		$maxEntityVisits
+	) {
+		$this->entityLookup = $entityLookup;
+		$this->entityPrefetcher = $entityPrefetcher;
+		$this->maxDepth = $maxDepth;
+		$this->maxEntityVisits = $maxEntityVisits;
+	}
+
+	/**
+	 * Get the closest entity (out of $toIds), from a given entity. The starting entity, and
+	 * the target entities are (potentially indirectly, via intermediate entities) linked by
+	 * statements with the given property ID, pointing from the starting entity to one of the
+	 * target entities.
+	 *
+	 * @since 3.10
+	 *
+	 * @param EntityId $fromId
+	 * @param PropertyId $propertyId
+	 * @param EntityId[] $toIds
+	 *
+	 * @return EntityId|null Returns null in case none of the target entities are referenced.
+	 * @throws ReferencedEntityIdLookupException
+	 */
+	public function getReferencedEntityId( EntityId $fromId, PropertyId $propertyId, array $toIds ) {
+		if ( !$toIds ) {
+			return null;
+		}
+
+		$this->alreadyVisited = [];
+
+		$steps = $this->maxDepth + 1; // Add one as checking $fromId already is a step
+		$toVisit = [ $fromId ];
+
+		while ( $steps-- ) {
+			$this->entityPrefetcher->prefetch( $toVisit );
+			$toVisitNext = [];
+
+			foreach ( $toVisit as $curId ) {
+				$result = $this->processEntityById( $curId, $fromId, $propertyId, $toIds, $toVisitNext );
+				if ( $result ) {
+					return $result;
+				}
+			}
+			// Remove already visited entities
+			$toVisit = array_unique(
+				array_diff( $toVisitNext, array_keys( $this->alreadyVisited ) )
+			);
+
+			if ( !$toVisit ) {
+				return null;
+			}
+		}
+
+		// Exhausted the max. depth without finding anything.
+		throw new MaxReferenceDepthExhaustedException(
+			$fromId,
+			$propertyId,
+			$toIds,
+			$this->maxDepth
+		);
+	}
+
+	/**
+	 * Find out whether an entity (directly) references one of the target ids.
+	 *
+	 * @param EntityId $id Id of the entity to process
+	 * @param EntityId $fromId Id this lookup started from
+	 * @param PropertyId $propertyId
+	 * @param EntityId[] $toIds
+	 * @param EntityId[] &$toVisit List of entities that still need to be checked
+	 * @return EntityId|null Target id the entity refers to, null if none.
+	 */
+	private function processEntityById(
+		EntityId $id,
+		EntityId $fromId,
+		PropertyId $propertyId,
+		array $toIds,
+		array &$toVisit
+	) {
+		$entity = $this->getEntity( $id, $fromId, $propertyId, $toIds );
+		if ( !$entity ) {
+			return null;
+		}
+
+		$mainSnaks = $this->getMainSnaks( $entity, $propertyId );
+
+		foreach ( $mainSnaks as $mainSnak ) {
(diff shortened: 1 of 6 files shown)


tests/unit/Lookup/EntityRetrievingClosestReferencedEntityIdLookupTest.php
```diff
@@ -0,0 +1,476 @@
...
+ * @author Marius Hoch
+ */
+class EntityRetrievingClosestReferencedEntityIdLookupTest extends PHPUnit_Framework_TestCase {
+
+	/**
+	 * @param EntityLookup $entityLookup
+	 * @param int $expectedNumberOfGetEntityCalls
+	 * @return EntityLookup
+	 */
+	private function maskEntityLookup( EntityLookup $entityLookup, $expectedNumberOfGetEntityCalls ) {
```
JeroenDeDauw: Is `mask` a typo? If not, I am confused by this verb.

tests/unit/Lookup/EntityRetrievingClosestReferencedEntityIdLookupTest.php
```diff
@@ -0,0 +1,476 @@
...
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ *
+ * @license GPL-2.0-or-later
+ * @author Marius Hoch
+ */
+class EntityRetrievingClosestReferencedEntityIdLookupTest extends PHPUnit_Framework_TestCase {
+
+	/**
+	 * @param EntityLookup $entityLookup
+	 * @param int $expectedNumberOfGetEntityCalls
```
JeroenDeDauw: Why have this argument an an if-else? Looks like you can just have two simple methods

tests/unit/Lookup/EntityRetrievingClosestReferencedEntityIdLookupTest.php
```diff
@@ -0,0 +1,476 @@
...
+		}
+
+		return $entityLookup;
+	}
+
+	/**
+	 * @param int $expectedPrefetches
+	 * @return EntityPrefetcher
+	 */
+	private function getEntityPrefetcher( $expectedPrefetches ) {
```
JeroenDeDauw: In the fundraising codebase we use the convention to name such methods in tests that construct new stuff `newStuff` and not `getStuff`. I think that is clearer, since then you know you get a new instance every time and can happily cause changes in its state without other stuff blowing up

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
```diff
@@ -0,0 +1,236 @@
...
+			$this->entityPrefetcher->prefetch( $toVisit );
+			$toVisitNext = [];
+
+			foreach ( $toVisit as $curId ) {
+				$result = $this->processEntityById( $curId, $fromId, $propertyId, $toIds, $toVisitNext );
+				if ( $result ) {
+					return $result;
+				}
+			}
+			// Remove already visitted entities
```
lucaswerkmeister: typo: visited

tests/unit/Lookup/EntityRetrievingClosestReferencedEntityIdLookupTest.php
```diff
@@ -0,0 +1,476 @@
...
+			new Item( $q90, null, null, $this->getReferencingStatementList( $pSubclassOf, [ $q3 ] ) )
+		);
+		$lookup->addEntity( new Item( $q1234, null, null, null ) );
+		$lookup->addEntity( new Item( $q3, null, null, null ) );
+
+		return $lookup;
+	}
+
+	/**
+	 * @return InMemoryEntityLookup
```
lucaswerkmeister: Does anything else depend on this specifically returning an InMemoryEntityLookup and not any EntityLookup?

tests/unit/Lookup/EntityRetrievingClosestReferencedEntityIdLookupTest.php
thiemowmde: I'm also a fan of the `new…` naming convention.

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
```diff
@@ -0,0 +1,236 @@
...
+	 */
+	private $entityLookup;
+
+	/**
+	 * @var EntityPrefetcher
+	 */
+	private $entityPrefetcher;
+
+	/**
+	 * @var int Maximum search depth: Maximum number of intermediate entities to search through.
```
thiemowmde, line 38: Oh, this is confusing. A "maximum depth" is not the same as a "maximum number of entities to search through". This search spans a tree. If the tree is very flat, the search depth might be something like 5, while a few thousand entities have been traversed already. Please count the entities, and not the depth. We run into issues with this in the QualityConstrains codebase already. @lucaswerkmeister might be able to provide an example.

Further down in the constructor I see both a depth and some maximum number. It's hard to tell what they do. I think a depth is not needed.

If you think having both is helpful, I would not make the depth throw an exception. An exception is only thrown if the maximum number of entities is reached. My reasoning for this idea is: Think of a tree with two branches, the first is extremely deep, while the second is only a few entities deep. Both together exceed the "maximum number of entities" setting. If the depth-check throws an exception, nothing would be found. But if the depth-check only makes the code stop traversing a particular branch, the second, smaller branch would be found.

I hope this makes sense.

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
```diff
@@ -0,0 +1,236 @@
...
+	 *
+	 * @return EntityId|null Returns null in case none of the target entities are referenced.
+	 * @throws ReferencedEntityIdLookupException
+	 */
+	public function getReferencedEntityId( EntityId $fromId, PropertyId $propertyId, array $toIds ) {
+		if ( !$toIds ) {
+			return null;
+		}
+
+		$this->alreadyVisited = [];
```
thiemowmde, line 96: Couldn't you keep this cache for the lifetime of the object?

lucaswerkmeister (commented):
LGTM in general, not tested manually yet (but I guess that would have to wait until a Lua binding is available in some other change?)

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
```diff
@@ -0,0 +1,236 @@
...
+	/**
+	 * @param EntityId $id Id of the entity to get
+	 * @param EntityId $fromId Id this lookup started from
+	 * @param PropertyId $propertyId
+	 * @param EntityId[] $toIds
+	 *
+	 * @return StatementListProvider|null Null if not applicable.
+	 */
+	private function getEntity( EntityId $id, EntityId $fromId, PropertyId $propertyId, array $toIds ) {
+		$this->alreadyVisited[$id->getSerialization()] = true;
```
thiemowmde: I miss a safety check here that starts screaming when this is called with an already visited entity. As far as I can see this is not required, as the caller makes sure this can't happen. But this is a little hidden in an array_diff, and I would feel better with an additional isset check. It's cheap.

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
mariushoch: Done :)

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
mariushoch: I hope I correctly understood your comment, if not, please correct me!

Please note that this is not doing a _depth-first search_, but a _breadth-first search_, thus there's no need to make "the code stop traversing a particular branch". In this context I think it might be useful for a caller to say that they only want us to look for paths with a certain maximum length.
The entity visit limitation is just an added safety, which I don't imagine firing very often, but only in specific cases where one or more visited entities references many other entities.

I agree that the documentation for this might be improvable. I'll see what I can do there.

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
mariushoch: No, because we want to maybe visit the same entities again, if another (or even the same) reference-relationship is queried.
We may have a caching decorator for this in the future, but that's not this classes business.

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
mariushoch: Added a `LogicException` here.

mariushoch: Changed this to actually trigger a warning now… throwing an exception is maybe a bit harsh.

tests/unit/Lookup/EntityRetrievingClosestReferencedEntityIdLookupTest.php
mariushoch: True, can even be one simpler method… changed to just use phpunit mocks all the time and not sometimes rely on `RestrictedEntityLookup`.

tests/unit/Lookup/EntityRetrievingClosestReferencedEntityIdLookupTest.php
mariushoch: It was not, but I see that it's not accurate… so changed to `restrictEntityLookup`.

tests/unit/Lookup/EntityRetrievingClosestReferencedEntityIdLookupTest.php
mariushoch: Good idea, changed.

tests/unit/Lookup/EntityRetrievingClosestReferencedEntityIdLookupTest.php
mariushoch: No, changed the return type hint.

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
JeroenDeDauw: `trigger_error` causes a global side effect. Without knowing the context here I can't say what is appropriate, though normally you either throw an exception and let the caller deal with it, which might be ignoring the error, escalating it, logging it, etc. Or, in cases where the execution in the local context should not be aborted, you log something or otherwise output info via some collaborator (which could also be an event thing or whatever).

lucaswerkmeister:
One thing I just remembered – what’s the status of usage tracking in this service? We discussed this in #193, but I don’t see anything related in this pull request.

mariushoch:
> what’s the status of usage tracking in this service?

I discussed this with @lydiapintscher and we probably don't need usage tracking for now.

If we want usage tracking later on, we would need this service to track the specific patch chosen… or in this case, where the service even returns the shortest path, we would need to track all visited entities, if we wanted to keep this guarantee all the time.

src/Lookup/EntityRetrievingClosestReferencedEntityIdLookup.php
thiemowmde: I'm afraid I still don't know what this integer represents. The documentation uses the words "depth" (as in how deep a tree is) and "number" (as in the number of visited leaves) in the same sentence. For me these are two very different concepts. Which one does this integer represent?

lucaswerkmeister: My understanding is that this represents the maximum *depth* of the tree. You can also think of the depth as the *number of entities above the current leaf* (i. e., intermediate entities) at any point. Does that make sense?

thiemowmde: > You can also think of the depth as the number of entities above the current leaf […]

"Above the current leave" is an other tree. What "number of entities" from this tree is meant? A horizontal or a vertical slice?

mariushoch: `Maximum number of intermediate entities to search through`, this means given we have a path `Q1 -> … -> Q42`, this is the number of entities represented by the `…`.

thiemowmde: So "maxDepth" is indeed the maximum depth of that particular tree. :-) Thanks!

PR #199, opened 2018-08-30
No longer spcify compat with Diff 1.x

JeroenDeDauw:


thiemowmde:
What's the reasoning? There is no commit message other than the first line, so I can't tell. Is it in fact not compatible any more? If so, which change broke the compatibility?

manicki:
Diff 1.x seems quite prehistory, so I guess we don't gain much by maintaining this compatibility.
Is there anything that still uses 1.x line of that lib?

manicki (approved):

JeroenDeDauw:
I noticed that this component specifies ~2.3|~1.0, which is a bit weird of a range and made me wonder if 1.0 even still works. No reason to spend effort on finding that out though since we indeed don't gain anything by still supporting it.

PR #202, opened 2018-11-06
Update Wikibase PHPCS rule set to latest release

thiemowmde:
The only change is an update to all `@license` tags, to match the new SPDX version.

~~I'm also removing the ArrayBracketSpacing sniff from this local rule set, because the code style enforced by this sniff (it *forbids* to use spaces in `$array[$key]`) clashes with the MediaWiki core code style that suggests to add these spaces.~~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
 	"require-dev": {
 		"phpmd/phpmd": "~2.3",
 		"phpunit/phpunit": "~5.7",
-		"wikibase/wikibase-codesniffer": "^0.3.0"
+		"wikibase/wikibase-codesniffer": "^0.5.0"
 	},
 	"autoload": {
 		"psr-4": {
--- a/src/ByPropertyIdGrouper.php
+++ b/src/ByPropertyIdGrouper.php
@@ -13,7 +13,7 @@ use Wikibase\DataModel\PropertyIdProvider;
  *
  * @since 1.0
  *
- * @license GPL-2.0+
+ * @license GPL-2.0-or-later
  * @author Bene* < benestar.wikimedia@gmail.com >
  */
 class ByPropertyIdGrouper {
--- a/src/DataValue/ValuesFinder.php
+++ b/src/DataValue/ValuesFinder.php
@@ -14,7 +14,7 @@ use Wikibase\DataModel\Snak\Snak;
  *
  * @since 1.1
  *
- * @license GPL-2.0+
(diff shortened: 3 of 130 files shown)


lucaswerkmeister:
> I'm also removing the ArrayBracketSpacing sniff from this local rule set, because the code style enforced by this sniff (it *forbids* to use spaces in `$array[$key]`) clashes with the MediaWiki core code style that suggests to add these spaces.

[This document](https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP) forbids spaces on array *access* (though they’re required in array *declaration*). Is there some other document or configuration that’s inconsistent with this?

thiemowmde:
I'm confused. It looks like [this is relatively new](https://www.mediawiki.org/w/index.php?diff=2227991). However: I think such a non-trivial detail of a coding style guide should be discussed and agreed on upstream, either in the MediaWiki or Wikibase rule set. A local exception is more problematic than helpful, in my opinion.

We can also remove this change from this patch set, if you prefer. Feel free to update the patch.

lucaswerkmeister:
Relatively new? I guess that’s a matter of perspective – for me, this change from September 2016 has always been there, since before I started ;)

And I don’t recall encountering much PHP code that puts spaces inside array access for any of the MediaWiki-related code I’ve worked on. (It’s different for JavaScript – we typically use the spaces there. Meh.)

JeroenDeDauw:
MW PHP CS has been like this forever:

```php
$a = [ 1, 2 ];
$b = $a[0];
```

thiemowmde:
Guys, I'm not fighting this, the contrary. Personally I'm very much a fan of exactly this style. But still: I suggest to enable it in the Wikibase rule set, or suggest it for the upstream MediaWiki one.

thiemowmde:
I reverted the line now, because it has nothing to do with what this patch originally was about.

lucaswerkmeister (approved):

lucaswerkmeister:
> I suggest to enable it in the Wikibase rule set, or suggest it for the upstream MediaWiki one.

Sounds good to me. I’m currently checking how many violations that reports for MediaWiki.

lucaswerkmeister:
333 violations, apparently (probably closer to ~165 in reality, since `$x[ 'a' ]` reports two violations, one for each space). But I also got some unrelated violations, so I’m not sure how much MediaWiki sticks to `phpcs` anyways?

PR #204, opened 2018-12-03
Make it clear I am not the only author

JeroenDeDauw:


JeroenDeDauw:
Tests passed, only hvvm failed for unrelated reasons

composer.json
```diff
@@ -14,6 +14,9 @@
 			"name": "Jeroen De Dauw",
 			"email": "jeroendedauw@gmail.com",
 			"homepage": "https://www.entropywins.wtf/"
+		},
+		{
+			"name": "Wikidata Team"
```
thiemowmde: The existing pattern in other codebases goes like this:
```JSON
{
	"name": "The Wikidata team",
	"role": "Developer"
}
```
Can we follow this here to stay consistent?

JeroenDeDauw: Totally. Please use the edit button

Ladsgroup: Should we fix issues in your pull request? Is it the new process of code review?

thiemowmde (approved):

PR #205, opened 2018-12-03
Add vararg constructor to InMemoryEntityLookup

JeroenDeDauw:
Noticed absence in
https://gerrit.wikimedia.org/r/c/mediawiki/extensions/WikibaseQualityConstraints/+/473719/1/tests/phpunit/Helper/ExceptionIgnoringEntityLookupTest.php

### DIFF
--- a/src/Lookup/InMemoryEntityLookup.php
+++ b/src/Lookup/InMemoryEntityLookup.php
@@ -31,6 +31,15 @@ class InMemoryEntityLookup implements EntityLookup {
 	private $exceptions = [];
 
 	/**
+	 * @param EntityDocument ...$entities
+	 */
+	public function __construct( ...$entities ) {
+		foreach ( $entities as $entity ) {
+			$this->addEntity( $entity );
+		}
+	}
+
+	/**
 	 * @param EntityDocument $entity
 	 *
 	 * @throws InvalidArgumentException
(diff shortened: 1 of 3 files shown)


src/Lookup/InMemoryEntityLookup.php
```diff
@@ -30,6 +30,15 @@ class InMemoryEntityLookup implements EntityLookup {
 	 */
 	private $exceptions = [];
 
+	/**
+	 * @param EntityDocument ...$entities
+	 */
+	public function __construct( ...$entities ) {
```
thiemowmde, line 36: I'm somewhat surprised by this. I faced heavy backlash when I tried to introduce pretty much the exact same back then in https://github.com/wmde/WikibaseDataModel/pull/376 and https://github.com/wmde/WikibaseDataModel/pull/416. Not meant to block this, not even as a vote. I'm just curious what changed and what others think nowadays.

JeroenDeDauw: What I am doing here is adding an array parameter using varargs. That is not the same as changing an existing parameter to support two input styles. The context is also different.

thiemowmde: I would not call this an "array parameter". The `...` syntax as documented at http://php.net/manual/en/functions.arguments.php#functions.variable-arg-list does not allow callers to pass an array, but a "variable-length list of arguments". Which is what I wanted in these other pull requests.

JeroenDeDauw:
@manicki can you get this simple change unstuck?

mariushoch:
+1 Looks good to me

manicki (approved):

PR #206, opened 2018-12-03
Drop hhvm from travis because broken

JeroenDeDauw:
All new PRs are red now:
* https://github.com/wmde/WikibaseDataModelServices/pull/205
* https://github.com/wmde/WikibaseDataModelServices/pull/204

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ matrix:
   fast_finish: true
   include:
     - env: DM=~6.3
-      php: hhvm
-    - env: DM=~7.5
       php: 5.6
     - env: DM=~7.5
       php: 7


Ladsgroup:
WMF is still on hhvm, I suggest not dropping it until WMF has migrated to php7

JeroenDeDauw:
It's either dropping this or fixing the issue. I think fixing the issue would be a waste of time. Don't let me stop you though. If you think you can justify spending time on it, by all means go for it.

thiemowmde:
> […] because broken

What exactly is broken? What would be needed to fix it properly?

I had a brief look at the error messages in #205 and it seems the issue might as well be fixable by temporarily disabling/removing PHPMD.

thiemowmde:
The PHPMD idea seems to work, see #207.

PR #210, opened 2019-01-22
Update wikimedia/assert, minimum PHP version

tgr:
Upgrade wikimedia/assert to 0.4.0 (no major changes but needed to avoid incompatible dependencies in mediawiki/vendor). Upgrade PHP requirement to 7.x or HHVM (needed by wikimedia/assert 0.4 and the end-of-life for 5.6 was January 1, anyway).

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,4 +1,7 @@
-build: true
+build:
+    environment:
+        php: 7.2
+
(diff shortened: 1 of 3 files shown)


tgr:
See https://github.com/wmde/WikibaseDataModel/pull/807#issuecomment-456227675

PR #211, opened 2019-02-01
Move FingerprintPatcher out of Internal namespace

lucaswerkmeister:
This class is needed to write patchers for non-Item/Property entities with fingerprints, so let’s make it public. Also, like other patchers in the library it now accepts any Diff, not just an EntityDiff; this makes it more useful to patch fingerprints not connected to proper entities, and also reduces the dependency upon EntityDiff’s label / description / aliases getters, which are already marked with a FIXME because not all entities have labels, descriptions or aliases.

---

Note: add `-M42` to Git commands (`status`, `diff` etc.) to make it recognize that `FingerprintPatcher.php` was moved, not removed and added anew.

Context is [I2e68f1a6dd](https://gerrit.wikimedia.org/r/487150) and [I8ceb9e6005](https://gerrit.wikimedia.org/r/487152).


src/Diff/FingerprintPatcher.php
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Wikibase\DataModel\Services\Diff;
+
+use Diff\DiffOp\Diff\Diff;
+use Diff\Patcher\PatcherException;
+use Wikibase\DataModel\Services\Diff\Internal\AliasGroupListPatcher;
+use Wikibase\DataModel\Term\Fingerprint;
+
+/**
+ * @since 1.0
```
thiemowmde: I think this needs to say `@since 3.13` now, or be removed.

src/Diff/FingerprintPatcher.php
```diff
@@ -0,0 +1,69 @@
...
+	 */
+	private $aliasGroupListPatcher;
+
+	public function __construct() {
+		$this->termListPatcher = new TermListPatcher();
+		$this->aliasGroupListPatcher = new AliasGroupListPatcher();
+	}
+
+	/**
+	 * @since 1.0
```
thiemowmde: Can you remove this `@since`?

thiemowmde (approved):
Good for me.

src/Diff/FingerprintPatcher.php
lucaswerkmeister: Yeah, I wasn’t sure what do to about these… I guess 3.13 makes more sense, yeah.

JeroenDeDauw:
This chance looks fine to me locally, though as per my email, I don't think it is a good idea to use this in Wikibase Schema.

mariushoch:
@lucaswerkmeister I guess this can be closed now?

lucaswerkmeister:
I guess so, though no one has objected to this patch itself…

PR #212, opened 2019-02-06
3.13 release

tgr:
New release with the upped PHP version requirement that makes this library compatible with the new release of wikimedia/assert.


Ladsgroup:
Aand relased. Thank you!

PR #213, opened 2019-02-20
Document IllegalArgumentException as throwable in EntityIdFormatter

AlaaSarhan:
Bug: https://phabricator.wikimedia.org/T207479

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
   fast_finish: true
   include:
     - env: DM=~6.3
-      php: hhvm-3.30
+      php: hhvm-3.18
     - env: DM=~7.5
       php: 7
     - env: DM=~8.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
   fast_finish: true
   include:
     - env: DM=~6.3
-      php: hhvm
+      php: hhvm-3.30
     - env: DM=~7.5
       php: 7
     - env: DM=~8.0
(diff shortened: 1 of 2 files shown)


src/EntityId/EntityIdFormatter.php
```diff
@@ -17,6 +17,7 @@ interface EntityIdFormatter {
 	 * @param EntityId $value
 	 *
 	 * @return string Either plain text, wikitext or HTML.
+	 * @throws \IllegalArgumentException
```
thiemowmde: Can you please add half a sentence **when** this exception is thrown?

.travis.yml
```diff
@@ -4,7 +4,7 @@ matrix:
   fast_finish: true
   include:
     - env: DM=~6.3
-      php: hhvm
+      php: hhvm-3.30
```
thiemowmde: This change appears to be unrelated. It's not mentioned in the commit message or the Phabricator ticket.

AlaaSarhan: It happens that current hhvm nightly does not support current version of composer (and soon enough not php). Travis was failing due to that.

This can be done in a separate PR, but think it's fine for this library to be done here since it's done in a separate commit.

thiemowmde: Sure, I just wanted to have the reason documented. :-)

src/EntityId/EntityIdFormatter.php
```diff
@@ -17,6 +17,7 @@ interface EntityIdFormatter {
 	 * @param EntityId $value
 	 *
 	 * @return string Either plain text, wikitext or HTML.
+	 * @throws \IllegalArgumentException when value falls out of supported patterns by the formatter
```
JeroenDeDauw: Not on a dev machine so I can't check - should this not be InvalidArgumentException? That is what the phab ticket says

AlaaSarhan: yes my bad .. fixed now

PR #219, opened 2019-05-14
Add FingerprintDiffer

JeroenDeDauw:
For optimization in https://phabricator.wikimedia.org/T220150

This is a re-submit of https://github.com/wmde/doctrine-term-store/pull/9

### DIFF
--- a/src/Diff/FingerprintDiffer.php
+++ b/src/Diff/FingerprintDiffer.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Diff;
+
+use Wikibase\DataModel\Term\AliasGroup;
+use Wikibase\DataModel\Term\Fingerprint;
+
+/**
+ * @license GPL-2.0-or-later
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class FingerprintDiffer {
+
+	/**
+	 * @param Fingerprint $first
+	 * @param Fingerprint $second
+	 *
+	 * @return Fingerprint Contains the terms present in $first that are not present in $second
+	 */
+	public function diff( Fingerprint $first, Fingerprint $second ): Fingerprint {
+		$difference = new Fingerprint();
+
+		foreach ( $first->getLabels() as $term ) {
+			if ( !$second->getLabels()->hasTerm( $term ) ) {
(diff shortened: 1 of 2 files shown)


lucaswerkmeister (changes requested):
This has an extremely confusing name. Under `FingerprintDiffer`, I expect a class that can be used by `ItemDiffer` and `PropertyDiffer` (partially replacing their current `toDiffArray` methods, just like they already use a `StatementListDiffer` to diff claims), accepting two `Fingerprint`s and returning some subclass of `Diff`, which can then be passed into `FingerprintPatcher` (which exists!). As far as I can tell, this implements something very different and not as generally useful (due to the asymmetry between `$first` and `$second`). I’m not sure whether that more specific class should be in data-model-services (it sounds like it’s fairly specific to the term store use-case), but I’m certain that it should not be called a `FingerprintDiffer`.

JeroenDeDauw:
This class follows the array_diff pattern. What would you call it?

Also, we have one use case where this is better for the consumer than using the Diff library and zero use cases where the opposite is true.

lucaswerkmeister:
In [Add `SchemaPatcher` class](https://gerrit.wikimedia.org/r/c/mediawiki/extensions/WikibaseSchema/+/487152), I used a `FingerprintPatcher`; if a `FingerprintDiffer` existed, I’m sure I would have tried to use it in [Add `SchemaDiffer` class](https://gerrit.wikimedia.org/r/c/mediawiki/extensions/WikibaseSchema/+/487151) as well, and gotten frustrated when it doesn’t match the rest of the library at all.

Ideas for alternative names: something with “fingerprint removal”,  or “complement” instead of “difference”.

JeroenDeDauw:
The current name is consistent with array_diff and the other diff functions in PHP. You're right that it uses a different approach then the Diff library based code. Not everything needs to follow the same approach.

For the use case you linked it might well have been good to have something based on Diff, though that is not the use case this code is written for. If you get frustrated with some code written for a different use case not serving your use case, that is not a problem with the code.

I cannot think of a more sensible name than the one used here and no concrete alternative has been suggested so far.

mariushoch:
Maybe name this `DifferenceFingerprint…` (where `…` something like `Service`, or `Computer`, …) given that this outputs the difference of two `Fingerprint`s in a `Fingerprint`?

PR #220, opened 2019-07-12
Add FallbackLabelDescriptionLookup interface

lucaswerkmeister:
This is an interface for LabelDescriptionLookup implementations that always return TermFallback instances, since some code likes to rely on the more specific return type. It was originally introduced in Wikibase change [Ie39f06bb33][1] and is now copied into this library.

[1]: https://gerrit.wikimedia.org/r/520902

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel Services release notes
 
+## Version 3.16.0 (dev)
+
+* Added `FallbackLabelDescriptionLookup` interface
+
 ## Version 3.15.0 (2019-04-24)
 
 * Added `ItemLookup` implementations
--- a/src/Lookup/FallbackLabelDescriptionLookup.php
+++ b/src/Lookup/FallbackLabelDescriptionLookup.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Lookup;
+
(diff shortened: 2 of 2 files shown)


JeroenDeDauw:
+1

lucaswerkmeister:
(The [force-push](https://github.com/wmde/WikibaseDataModelServices/compare/d437a8def10cf2963b48bf5d22654e86d51861bc..f0a1aa54bc83b6c665db1011de22df9fe5610e2c) only fixes a typo in the commit message.)

RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel Services release notes
 
+## Version 3.16.0 (dev)
```
AlaaSarhan, line 3: Just to confirm my understanding..
This is an interesting process .. so we add stuff under this that stays at `dev` until the actual release is made and then date would be put in here as a commit for the release as well right?

AlaaSarhan (approved):

RELEASE-NOTES.md
JeroenDeDauw: This is what I have been doing since 2009 or so. What's interesting about this / what would you do instead? :)

AlaaSarhan: oh no I wouldn't do anything instead.. I find it interesting in a positive way.. I like it and would keep doing it.

PR #221, opened 2019-07-25
Follow the chain in the double redirects

Ladsgroup:
Double redirects are quite a common thing in Wikimedia, refusing
to resolve them will cause lots of production fatals.

Bug: [T228996](https://phabricator.wikimedia.org/T228996)

### DIFF
--- a/src/Lookup/RedirectResolvingEntityLookup.php
+++ b/src/Lookup/RedirectResolvingEntityLookup.php
@@ -86,9 +86,9 @@ class RedirectResolvingEntityLookup implements EntityLookup {
 
 			// Try one more time
 			try {
-				$this->lookup->hasEntity( $ex->getRedirectTargetId() );
+				return $this->lookup->hasEntity( $ex->getRedirectTargetId() );
 			} catch ( UnresolvedEntityRedirectException $ex2 ) {
-				$this->lookup->hasEntity( $ex2->getRedirectTargetId() );
+				return $this->lookup->hasEntity( $ex2->getRedirectTargetId() );
 			}
 		}
 	}
--- a/src/Lookup/RedirectResolvingEntityLookup.php
+++ b/src/Lookup/RedirectResolvingEntityLookup.php
@@ -50,7 +50,17 @@ class RedirectResolvingEntityLookup implements EntityLookup {
 		try {
 			return $this->lookup->getEntity( $entityId );
 		} catch ( UnresolvedEntityRedirectException $ex ) {
-			return $this->lookup->getEntity( $ex->getRedirectTargetId() );
+			// Avoid self-redirects
+			if ( $ex->getRedirectTargetId() === $entityId ) {
+				throw $ex;
+			}
+
+			// Try one more time
+			try {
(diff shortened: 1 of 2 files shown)


AlaaSarhan (approved):

manicki:
Understanding question:
@Ladsgroup  you wrote above that
> Double redirects are quite a common thing in Wikimedia

whereas the linked phabricator task includes the comment https://phabricator.wikimedia.org/T228996#5365525 saying
> I don’t think Wikibase should follow the double redirect – that would go against longstanding Wikipedia tradition

and links to the English Wikipedia page https://en.wikipedia.org/wiki/Wikipedia:Double_redirects saying in its first paragraph:
> A redirect that points to another redirect is called a double redirect. These pages are unwanted

Is there a contradiction between those statements or is it all fine?

AlaaSarhan:
> Is there a contradiction between those statements or is it all fine?

there seems to be a contradiction indeed in the motivations at least. Though, the document says that double-redirects are just unwanted (not forbidden by the software) as there are seemingly ways to achieve it as I mentioned https://phabricator.wikimedia.org/T228996#5367228

The wikipedia page also describes the case of moving a page that was the destination of a redirect to be one source of double redirects.

----

As long as the software cannot prohibit creating those double-redirects, nor it resolves them statically when they are about to happen, it should probably at least handle resolving them as one might expect (at least without a fatal). The two options appear to be either to resolve the double-redirect as if it was one redirect (like this PR tries to do) or to render the redirection page after the first one, like the wikipedia article on double-redirects mentions (example on wikidata with lexemes https://wikidata.beta.wmflabs.org/w/index.php?title=Lexeme:L3338&redirect=no).

----

Ladsgroup:
> Is there a contradiction between those statements or is it all fine?

At first it might seems so but if we dive deep, no. Yes, double redirects are undersirable and they should be fixed (in fact, I improved the pywikibot script that fixes double redirects to support wikibase entities a couple of years ago) but they happen quite a lot.

That being said, I'm all fine with moving to have only graceful degradation instead of resolving them. It seems it has been intentionally not being resolved in this code.

AlaaSarhan (changes requested):
to avoid merge by mistake. We agreed to go with a different approach, graceful handling of the issue within Lexeme instead of implementing double-redirect in low-level services

lucaswerkmeister:
MediaWiki has [a setting](https://www.mediawiki.org/wiki/Manual:$wgMaxRedirects) for how many redirects to follow, we definitely shouldn’t unconditionally follow more redirects than that. And in MediaWiki core that setting has apparently been [broken](https://phabricator.wikimedia.org/T32359) for years (since the Bugzilla days), and nobody seems to care much, because we don’t use `$wgMaxRedirects > 1` in production as far as I’m aware.

AlaaSarhan:
Thanks for the insights @lucaswerkmeister yeah seems it isn't used in production as it isn't set to any specific value when I search in mediawiki-config. So it must be using the default of 1 in all of our environments.

mariushoch:
@Ladsgroup Do you want to close this?

Ladsgroup:
> @Ladsgroup Do you want to close this?

Sure

Issue #222, opened 2019-11-15
Allow diff/diff ^3.0/3.2


reedy:
To be able to use Diff ^3.0 or ^3.2 in Wikimedia Production https://github.com/wmde/WikibaseDataModelServices/blob/master/composer.json#L30 needs updating/loosening :)

Which also means dropping PHP < 7.0 support

PR #223, opened 2019-11-15
Bump diff/diff to ~3.2 and php >= 7.0

reedy:
Ping https://github.com/wmde/WikibaseDataModelServices/issues/222

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,10 +24,10 @@
 		"irc": "irc://irc.freenode.net/wikidata"
 	},
 	"require": {
-		"php": ">=5.6.99",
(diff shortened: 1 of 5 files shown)


reedy:
Lets see what CI has to say ;D

reedy:
Guess hhvm should be dropped

composer.json
```diff
@@ -24,10 +24,10 @@
 		"irc": "irc://irc.freenode.net/wikidata"
 	},
 	"require": {
```
reedy, line 26: I imagine some of these old versions of the other libraries probably want dropping at some point

Issue #225, opened 2019-11-15
Make new release that uses diff ^3.2 (3.16.0?)


reedy:
Now https://github.com/wmde/WikibaseDataModelServices/issues/222 has landed. We need a release so we can use it

addshore:
https://phabricator.wikimedia.org/T241580

PR #228, opened 2020-01-06
Release 4.0.0

Ladsgroup:
Dropping support for hhvm and diff 2.3

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,8 +1,10 @@
 # Wikibase DataModel Services release notes
 
-## Version 3.16.0 (dev)
+## Version 4.0.0 (2020-01-06)
 
 * Added `FallbackLabelDescriptionLookup` interface
+* Stopped supporting hhvm and added support for php 7.3 and 7.4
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
3.17

mariushoch:
> 3.17

You mean 3.16 (last released is 3.15 AFAICT)? I'd also go for that, I don't think this needs a new major version.

Ladsgroup:
> > 3.17
> 
> You mean 3.16 (last released is 3.15 AFAICT)? I'd also go for that, I don't think this needs a new major version.

Yeah, it should be 3.16 but isn't dropping php version a breaking change?

jdforrester:
> > You mean 3.16 (last released is 3.15 AFAICT)? I'd also go for that, I don't think this needs a new major version.
> 
> Yeah, it should be 3.16 but isn't dropping php version a breaking change?

It is. We've done major versions in other libraries when dropping HHVM/PHP5.6/etc.

jdforrester:
(Also, +"Fixes #225.")

PR #229, opened 2020-01-23
Stop accepting null for term type and lang in TermBuffer::prefetchTerms()

Ladsgroup:
It has caused more harm than good. Lots of places load all terms to
use very small bits of it. They are fixed now.


addshore:
Going to need to end up fixing a bunch of usages with this change.
But I'm very pro being explicit here.

addshore (approved):

lucaswerkmeister (approved):

PR #234, opened 2020-02-12
Add interfaces for writing Item/Property term stores

mariushoch:
These tiny interfaces are meant to eventually replace `wikibase/term-store`.

### DIFF
--- a/src/Term/ItemTermStoreWriter.php
+++ b/src/Term/ItemTermStoreWriter.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Term;
+
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Term\Fingerprint;
+
+interface ItemTermStoreWriter {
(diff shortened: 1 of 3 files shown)


src/Term/ItemTermStoreWriter.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Wikibase\DataModel\Services\Term;
+
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Term\Fingerprint;
+
+interface ItemTermStoreWriter {
```
Ladsgroup, line 8: BTW. This is going to replace EntityTermStoreWriter interface

mariushoch: `EntityTermStoreWriter` is generally defined on EntityId/ EntityDocument, it might have a use for other entity types.

Ladsgroup (approved):
It looks good to me, I leave it to other to merge though.

PR #238, opened 2020-03-16
Simplify ByPropertyIdGrouper::addPropertyIdProvider()

lucaswerkmeister:
PHP automatically initializes arrays as empty when the `$array[] = $value` syntax is used to append to them, so there’s no need for the explicit `isset()` check or initialization.


wiese (approved):
I was suspecting a deeper reason for it be the way it was but [looking back at its initial introduction back in 2014](https://github.com/wmde/WikibaseDataModel/commit/fc5a158a6205bfbe7ee4c4da1f91dc5d4d89bc94#diff-dd0b59aee78953d4a34c512630c7db49R55) there never seems to have been more to it.

Ladsgroup (approved):

PR #242, opened 2020-11-11
Document language fallback behavior of interfaces

lucaswerkmeister:
Previously, it wasn’t clear whether these interfaces are supposed to apply language fallbacks or not. From the existing implementations in Wikibase, it seems that `TermLookup` and `TermBuffer` are intended only to look up terms in specific languages (without fallbacks), whereas `LabelLookup` and `LabelDescriptionLookup` may or may not apply fallbacks – implementations include `LanguageLabelDescriptionLookup` (no fallbacks) and `LanguageFallbackLabelDescriptionLookup` (fallbacks).

Bug: [T266145](https://phabricator.wikimedia.org/T266145)

### DIFF
--- a/src/Lookup/LabelDescriptionLookup.php
+++ b/src/Lookup/LabelDescriptionLookup.php
@@ -6,6 +6,11 @@ use Wikibase\DataModel\Entity\EntityId;
 use Wikibase\DataModel\Term\Term;
 
 /**
+ * Looks up the label or description of an entity.
+ *
+ * Like for {@link LabelLookup},
+ * implementations may or may not apply language fallbacks.
+ *
  * @since 1.1
  *
  * @license GPL-2.0-or-later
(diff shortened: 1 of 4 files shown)


chukarave:
@lucaswerkmeister what do you think about
"Not all implementations apply language fallbacks" instead of "Implementations may or may not apply fallbacks"?
Just to make the point a bit clearer.

lucaswerkmeister:
Hm, to me that sounds a bit like it’s expected that implementations apply language fallbacks, and there’s something wrong with those that don’t…

Maybe this?

> Whether language fallbacks are applied or not depends on the implementation.

I guess the underlying idea is that, for both “what language is the label in?” and “are language fallbacks applied?”, this is a decision made by whoever _created_ the service, and classes who _use_ the `LabelLookup` interface are supposed to trust that the instance they got injected has “the right” behavior, if that makes sense? But I’m not sure how to put that into words.

chukarave:
> > Whether language fallbacks are applied or not depends on the implementation.

I like this. I'm not sure it explains the whole background, but at least it makes it clear that not everything is built-in in the interface.

LGTM

chukarave:
It probably makes sense to change it in src/Lookup/LabelDescriptionLookup.php too?

lucaswerkmeister:
Indeed, thanks. (And I also forgot that we don’t force-push on GitHub, I think.)

PR #247, opened 2021-04-23
Extract EntityRedirectTargetLookup from EntityRedirectLookup

jakobw:
In some cases (see T280771) we only need to find redirect targets and
don't need EntityRedirectLookup::getRedirectIds. In order to not add
unnecessary implementations or violate the interface-segregation
principle EntityRedirectTargetLookup provides only the redirect ->
redirect target part of the interface.

Bug: T280771


tzhelyazkova (approved):
I think the commit message is missing the word "need" somewhere in there but the change is good to go otherwise 🚀

jakobw:
Thanks, I added the missing word :D

